PR #38, opened 2025-02-17
Previous and next buttons

CRogers:
## Before this PR
There are no previous and next buttons on each page.

## After this PR
There are previous and next buttons, as well as a link to the table of contents.

## Possible downsides?
Unfortunately, github markdown rendering is highly restrictive. We can't use inline styles at all so can't easily put the previous/next links aligned left and right at the same level using `display: flex` tricks. Instead, I've made a table.


CRogers:
:+1:

PR #43, opened 2025-02-19
Docs for Build Lifecycle and Avoiding Unnecessary Configuration

CRogers:
## Before this PR
We had no docs on these topics, and I have to explain what to do to people every time in code reviews.

## After this PR
I can just point people here.

In future: there will be errorprones that point people to these docs instead.


guide/avoiding-unnecessary-configuration.md
CRogers, line 1: You can go to links like this (or click "view file") to get rendered markdown where the links also work: https://github.com/palantir/gradle-guide/blob/callumr/lazy-instantiation/guide/avoiding-unnecessary-configuration.md

guide/avoiding-unnecessary-configuration.md
```diff
@@ -0,0 +1,161 @@
...
+});
+```
+
+This is better; we tell Gradle there is task called `foo`, but we give the configuration of the task lazily. If `foo` is not executed, we don't even realise or configure the task.
+
+## Using `TaskProvider`s
+
+The `register` approach now gives us a `TaskProvider`, which is a lazy handle to the task. Many  of the built-in Gradle APIs will let you pass a `TaskProvider` instead of a `Task` object.
+
+If you want to give use an output property of a task as the input to another task, you can just transform the `TaskProvider` into another `Provider` using `map` or `flatMap`. You can call `.configure` on a `TaskProvider` to lazily configure the task without realising it (avoid calling `.get()` on `TaskProvider`s).
```
felixdesouza: I believe you always want to use `flatMap`. `TaskProvider<Provider<X>>` doesn't always carry dependencies iirc

felixdesouza: I also think it's worth calling out the "avoid calling `.get()` on `TaskProvider`s" in a separate piece

CRogers: Yeah, I might leave that to the "task dependencies" section - or some other place about how task deps get removed. I'm also writing errorprones.

CRogers: >I believe you always want to use flatMap. TaskProvider<Provider<X>> doesn't always carry dependencies iirc

guide/avoiding-unnecessary-configuration.md
```diff
@@ -0,0 +1,161 @@
...
+project.getTasks().whenObjectAdded(task -> ...);
+```
+
+These will realise _all_ tasks, even ones already registered:
+
+```java
+project.getTasks().all(task -> ...);
+project.getTasks().matching(task -> ...);
+project.getTasks().withType(Foo.class).matching(task -> ...);
+```
```
felixdesouza: I think you also want to note the common pitfall of `withType(Foo.class, task -> {})` whilst they may look the same, this version is the equivalent of calling `withType().all()`. There are a number of different methods similar to this on the `TaskCollection`/`DomainObjectCollection` class hierarchy.

guide/avoiding-unnecessary-configuration.md
```diff
@@ -0,0 +1,161 @@
...
+project.getTasks().configureEach(task -> ...)
+```
+
+It's a bit like running `.configure` on each `TaskProvider` without realising the tasks.
+
+To move from `matching`, you can just use an `if` statement inside the `configureEach` rather than prefiltering the tasks.
+
+## Lazy `Configuration` registration
+
+`Configuration` objects should also be lazily registered. This follows the same pattern as tasks above. For example, you `register` rather than `create` a `Configuration`:
```
felixdesouza, line 143: is this strictly true? I wasn't aware of any perf penalty from realising `Configuration`s?

guide/avoiding-unnecessary-configuration.md
```diff
@@ -0,0 +1,161 @@
...
+
+// Later
+String configurationName = configuration.getName();
+```
+
+All of the `getByName`/`all` etc methods [described above for Tasks](#accessing-tasks-lazily) also apply for `Configuration`s.
+
+## Other types
+
+The advice for `Task`s also applies to other Gradle domain objects like `SourceSetContainer`, `ArtifactRepositoryContainer` - anything that inherits from [DomainObjectCollection](https://docs.gradle.org/current/javadoc/org/gradle/api/DomainObjectCollection.html#:~:text=All%20Known%20Subinterfaces%3A).
```
felixdesouza:
```suggestion
The advice for `Task`s also applies to other Gradle domain objects like `SourceSetContainer`, `ArtifactRepositoryContainer` - anything that inherits from [`DomainObjectCollection`](https://docs.gradle.org/current/javadoc/org/gradle/api/DomainObjectCollection.html#:~:text=All%20Known%20Subinterfaces%3A).
```

guide/lifecycle-of-a-gradle-build.md
```diff
@@ -0,0 +1,80 @@
...
+
+When you _invoke_ gradle, you call the `./gradlew` script and provide a set of "requested tasks":
+
+```
+./gradlew taskA taskB
+```
+
+This `./gradlew` script runs the "Gradle Wrapper" - a new process that runs code in the jar checked into your repo at `gradle/wrapper/gradle-wrapper.jar`. The Gradle Wrapper will ensure the Gradle distribution listed in `gradle/wrapper/gradle-wrapper.properties` is installed - if it isn't, it will download and install it. [Further reading about the Gradle Wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html).
+
+The Gradle Wrapper will now check to see if a "Gradle Daemon" already exists. Daemons are long-lived instances of Gradle which speed up build times by keeping the JVM running between builds. If a Daemon already exists, it will be reused. If not, a new one will be started. [Further reading about the Gradle Daemon](https://docs.gradle.org/current/userguide/gradle_daemon.html).
```
felixdesouza: is it worth talkign about daemon compatibility? i.e. you have daemons available but different jvm or some other weird thing is stopping it from being reused

felixdesouza:
👍  pending comments

guide/avoiding-unnecessary-configuration.md
CRogers: There should be about the same cost as configuring a `Task`, no? Which is probably very small, but worth avoiding. Configuring lazily has other benefits too, like producing a `Provider` which people can map rather than doing something like `project.provider(() -> configuration.something()).map()` that might not bring along task deps (I think?). I think it's also easier to just people in the habit of using providers for everything.

guide/avoiding-unnecessary-configuration.md
CRogers: ooo, this is a sneaky one I didn't know about. It's even in the gradle docs just non-obvious (they say to avoid just [`DomainObjectCollection.withType()`](https://docs.gradle.org/current/javadoc/org/gradle/api/DomainObjectCollection.html#withType-java.lang.Class-org.gradle.api.Action-).

felixdesouza (approved):

PR #58, opened 2025-03-10
Initial errorprones for lazy task creation

CRogers:
## Before this PR
There are no errorprone checks for good gradle.

## After this PR
==COMMIT_MSG==
Add `ConfigurationAvoidanceRegistration` error prone check.
==COMMIT_MSG==

I've added a check to error on `tasks.create` (or in general other `NamedDomainCollection.create`).

There's also adding the idea of "best effort" checks that are not usually run but can be run in say an excavator. The idea is to repeatedly run the best effort checks until nothing changes, and they will make tiny changes until the whole project is improved.


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
```diff
@@ -0,0 +1,136 @@
...
+            VisitorState state, SuggestedFix.Builder fixBuilder, VariableTree variableTree) {
+
+        boolean isTask = state.getTypes()
+                .isSubtype(ASTHelpers.getType(variableTree.getType()), state.getTypeFromString("org.gradle.api.Task"));
+
+        if (isTask) {
+            fixBuilder.addImport("org.gradle.api.tasks.TaskProvider");
+            fixBuilder.prefixWith(variableTree.getType(), "TaskProvider<");
+        } else {
+            fixBuilder.addImport("org.gradle.api.provider.Provider");
```
felixdesouza: do you not want this to be a `NamedDomainObjectProvider` ?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
```diff
@@ -0,0 +1,136 @@
...
+                state.getSourceForNode(ASTHelpers.getReceiver(tree.getMethodSelect())) + ".register");
+
+        if (parentPath.getLeaf() instanceof ExpressionTree || parentPath.getLeaf() instanceof ReturnTree) {
+            fix.postfixWith(tree, ".get()");
+        }
+
+        return description.addFix(fix.build()).build();
+    }
+
+    private static void replaceVariableDeclartionTypeWithProvider(
```
felixdesouza:
```suggestion
    private static void replaceVariableDeclarationTypeWithProvider(
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
```diff
@@ -0,0 +1,136 @@
...
+            replaceVariableUsagesWithTaskProviderGet(fix, parentPath);
+        }
+
+        fix.replace(
+                tree.getMethodSelect(),
+                state.getSourceForNode(ASTHelpers.getReceiver(tree.getMethodSelect())) + ".register");
+
+        if (parentPath.getLeaf() instanceof ExpressionTree || parentPath.getLeaf() instanceof ReturnTree) {
+            fix.postfixWith(tree, ".get()");
+        }
```
felixdesouza: any way you can replace this with a switch? a bit easier to see the control flow

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
```diff
@@ -0,0 +1,136 @@
...
+        SuggestedFixBuilder fix = new SuggestedFixBuilder();
+
+        if (!bestEffortModeEnabled(state)) {
+            return description.build();
+        }
+
+        // If the first argument is a map, or second is a Closure, there isn't an equivalent
+        // `.register` method to move to (from Java code at least)
+        if (NO_DIRECT_REGISTER_EQUIVALENT.matches(tree, state)) {
+            return description.build();
```
felixdesouza, line 78: should we still warn at the very least?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/besteffort/ProviderImprovements.java
```diff
@@ -0,0 +1,117 @@
...
+        summary = "Do not call `Provider.get`. Instead, pass providers directly to methods that accept them, "
+                + "or transform providers using `Provider.map` or `Provider.flatMap`, or combine providers using "
+                + "`Provider.zip`. Calling `Provider.get` causes Gradle to lose track of implicit dependencies and "
+                + "can lead to timing issues by reading values too early.")
+public final class ProviderImprovements extends GradleGuideBugChecker
+        implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final Matcher<ExpressionTree> MATCHER = Matchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.provider.Provider")
+            .namedAnyOf("get", "getOrNull", "getOrElse");
```
felixdesouza: not sure what you'd do in this instance, but I think `isPresent` is also something you want to avoid at configuration time where possible

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/besteffort/ProviderImprovementsTest.java
```diff
@@ -0,0 +1,250 @@
...
+                }
+                """,
+                    """
+                import org.gradle.api.provider.Provider;
+                import org.gradle.api.provider.ProviderFactory;
+
+                class Test {
+                    void test(ProviderFactory providerFactory, Provider<String> provider) {
+                        provider.map(providerValue -> providerValue + " yo");
+                    }
```
felixdesouza, line 75: hmm, do we have a thing that deletes unused variables?

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistrationTest.java
```diff
@@ -0,0 +1,292 @@
...
+                }
+            }
+            """);
+
+        compilationTestHelper.doTest();
+    }
+
+    @Test
+    void best_effort_refactors_usages_of_tasks_create_where_the_return_value_is_unused_to_register() {
+        bestEffortRefactoringValidator()
```
felixdesouza: can you split these into three separate calls to `bestEffortRefactoringValidator`

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistrationTest.java
```diff
@@ -0,0 +1,292 @@
...
+                    // BUG: Diagnostic contains: use `.register`
+                    tasks.create(Map.of("name", "lol", "type", Task.class), Closure.IDENTITY);
+                }
+            }
+            """)
+                .doTest();
+    }
+
+    @Test
+    void best_effort_refactors_usages_of_configurations_create_where_the_return_value_is_unused_to_register() {
```
felixdesouza: separate calls pls 🙏  I'm having to diff bigger chunks + scrolling => not fun

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
CRogers: Could be, I actually prefer using `Provider` unless you need the `.getName()` as it isn't so horribly verbose. That said, for best effort mode it'd probably be better as `NamedDomainObjectProvider` _if_ we're able to automatically make it use `.getName()` somehow.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
CRogers: This is "best effort" mode refactoring, so no point warning (I also think warnings are silly - they just clog up your compiler output and people rarely fix them). In regular compilation it will just make an actual error.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
CRogers: I don't think there's a nice way to do this without jdk 21 support.

CRogers: Without adding in casting at least

CRogers: Put the instanceofs next to each other at least

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistration.java
CRogers: Made it a NDOP, realised we probably need it for `.configure` as well.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/besteffort/ProviderImprovements.java
CRogers: This is vestigial from when I was actually erroring on `Provider.get/getOrNull` etc (which doesn't work without understanding the context it's being called it) - reduced it back to just `Provider.get` as that's what we handle so far.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/besteffort/ProviderImprovementsTest.java
CRogers: Eventually, it could be another rule in the best effort mode improvements. There might even be something in errorprone to do it.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistrationTest.java
CRogers: Ha - I don't know what I was thinking when I was doing this

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/ConfigurationAvoidanceRegistrationTest.java
CRogers: Done

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/besteffort/ProviderImprovementsTest.java
felixdesouza: I suppose, my actual question was, in this particular scenario, you will still have failing errorprones because now you have an unused variable, and the only way to move forward is to suppress, which is a bit sad

felixdesouza (approved):

PR #59, opened 2025-03-10
Delete unused code

CRogers:
## Before this PR
Old code that I forgot to delete is here

## After this PR
The code is deleted.


CRogers:
:+1:

PR #119, opened 2025-05-16
Expanded NonAbstractGradleType to add ability to detect if gradle managed properties are abstract and start with get

FinlayRJW:
## Before this PR
Had no Errorprone to check if task and extension properties where abstract and started with get - see https://github.com/palantir/gradle-idea-configuration/pull/20/files#r2084653438

## After this PR
Expanded `NonAbstractGradleType` to add ability to detect if gradle managed properties are abstract and start with get

==COMMIT_MSG==
Expanded NonAbstractGradleType to add ability to detect if gradle managed properties are abstract and start with get
==COMMIT_MSG==

### DIFF
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
@@ -91,9 +91,11 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
         if (enclosingClass == null) {
             return Description.NO_MATCH;
         }
+
         if (!IS_TASK.matches(enclosingClass, state)) {
             return Description.NO_MATCH;
         }
+
         if (!isManagedPropertyGetter(method, state)) {
             return Description.NO_MATCH;
         }
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
@@ -63,6 +63,10 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
     private static final Supplier<Type> CLASS_TYPE_SUPPLIER = Suppliers.typeFromString("java.lang.Class");
     private static final Supplier<Type> PROVIDER_TYPE_SUPPLIER =
             Suppliers.typeFromString("org.gradle.api.provider.Provider");
+    private static final Supplier<Type> DOMAIN_OBJECT_COLLECTION_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.DomainObjectCollection");
+    private static final Supplier<Type> FILE_COLLECTION_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.file.FileCollection");
 
     private static final String ABSTRACT_PROPERTY_METHOD_MSG = "Properties on Tasks or Extensions should be declared "
             + "abstract. Declare this method as 'public abstract', e.g., 'public abstract Property<Integer> "
@@ -138,7 +142,7 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
         }
 
         Type varType = ASTHelpers.getType(tree);
-        if (varType != null && isSubtypeOfProvider(varType, state)) {
+        if (varType != null && isManagedPropertyType(varType, state)) {
             return buildDescription(tree).setMessage(PROPERTY_FIELD_MSG).build();
         }
 
@@ -185,7 +189,7 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
         return StreamSupport.stream(extSym.members().getSymbols().spliterator(), false)
                 .filter(memberSym -> memberSym instanceof VarSymbol)
                 .map(memberSym -> (VarSymbol) memberSym)
-                .anyMatch(varSym -> isSubtypeOfProvider(varSym.type, state));
+                .anyMatch(varSym -> isManagedPropertyType(varSym.type, state));
     }
 
     private Description checkManagedPropertyGetters(ClassSymbol extSym, MethodInvocationTree tree, VisitorState state) {
@@ -223,12 +227,30 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
         return symbol != null && symbol.getModifiers().contains(Modifier.ABSTRACT);
     }
 
+    private static boolean isManagedPropertyType(Type type, VisitorState state) {
+        return isSubtypeOfProvider(type, state)
+                || isSubtypeOfDomainObjectCollection(type, state)
+                || isSubtypeOfFileCollection(type, state);
+    }
+
+    private static boolean isSubtypeOfProvider(Type type, VisitorState state) {
+        return ASTHelpers.isSubtype(type, PROVIDER_TYPE_SUPPLIER.get(state), state);
+    }
+
+    private static boolean isSubtypeOfDomainObjectCollection(Type type, VisitorState state) {
+        return ASTHelpers.isSubtype(type, DOMAIN_OBJECT_COLLECTION_TYPE_SUPPLIER.get(state), state);
+    }
+
+    private static boolean isSubtypeOfFileCollection(Type type, VisitorState state) {
+        return ASTHelpers.isSubtype(type, FILE_COLLECTION_TYPE_SUPPLIER.get(state), state);
+    }
+
     private static boolean isManagedPropertyGetter(MethodTree method, VisitorState state) {
         if (!method.getParameters().isEmpty()) {
             return false;
         }
         Type returnType = ASTHelpers.getType(method.getReturnType());
-        return returnType != null && isSubtypeOfProvider(returnType, state);
+        return returnType != null && isManagedPropertyType(returnType, state);
     }
 
     private static boolean isManagedPropertyGetter(MethodSymbol method, VisitorState state) {
@@ -236,11 +258,7 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
             return false;
         }
         Type returnType = method.getReturnType();
-        return returnType != null && isSubtypeOfProvider(returnType, state);
-    }
-
-    private static boolean isSubtypeOfProvider(Type type, VisitorState state) {
-        return ASTHelpers.isSubtype(type, PROVIDER_TYPE_SUPPLIER.get(state), state);
+        return returnType != null && isManagedPropertyType(returnType, state);
     }
 
     @Override
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
@@ -65,7 +65,7 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
             + "syntax work of the box.";
 
     private static final String PROPERTY_GETTER_NAMING_MSG =
-            "Gradle managed property getter methods must be named starting with 'get', e.g., 'getFoo'. This naming "
+            "Properties on Tasks or Extensions should be named starting with 'get', e.g., 'getFoo'. This naming "
                     + "convention is required for Gradle to recognize and manage the property, enabling automatic "
                     + "property wiring and Groovy DSL support, this will automatically this will make the `foo = 3` "
                     + "groovy syntax work of the box.";
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
@@ -1,3 +1,18 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -53,18 +58,48 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
...
 
-        if (tree.getModifiers().getFlags().contains(Modifier.ABSTRACT)) {
-            return Description.NO_MATCH;
-        }
-
         if (IS_TASK.matches(tree, state)) {
-            return buildDescription(tree).build();
+            if (!tree.getModifiers().getFlags().contains(Modifier.ABSTRACT)) {
+                return buildDescription(tree)
+                        .setMessage("Gradle managed Task types must be abstract classes.")
```
CRogers: This error message is not very descriptive. See the message in the `@BugPattern` this was before - it really should have plenty of information. I think for this one, we should just not set a custom message so it uses the default.

FinlayRJW: Ah I see I thought that was included in every message - will update now

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -53,18 +58,54 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
...
+            if (!tree.getModifiers().getFlags().contains(Modifier.ABSTRACT)) {
+                return buildDescription(tree).build();
+            }
+            return tree.getMembers().stream()
+                    .filter(member -> member instanceof MethodTree)
+                    .map(member -> (MethodTree) member)
+                    .filter(method -> isManagedPropertyGetter(method, state))
+                    .map(method -> {
+                        if (!method.getModifiers().getFlags().contains(Modifier.ABSTRACT)) {
+                            return buildDescription(method)
```
CRogers: This is problematic because if someone adds `@SuppressWarnings("NonAbstractGradleType")` on the method, it will not actually be suppressed. Error-prone handles not running your matcher (eg this ClassTreeMatcher) if the element is matching on has suppressed the check. But if you start descending the tree in your own check, you need to manually handle the suppressions. There are way to do this - you can use `SuppressibleTreePathScanner` to visit all the nodes below and will automatically handle suppressions, but this has the cost of visiting every single node in the tree in addition to the main error-prone visitor that calls all the match methods doing this.

A much more canonical way to do this is to actually implement `MethodTreeMatcher` in this class, then do these checks in `matchMethod`. error-prone will handle the suppressions for you. You'll need to check the enclosing class extends `Task` - I think there's something in `Matchers` to do this.

We need to do something similar to what you've written here for the Extensions as we have to do the check "outside" the extension class. But since we can do it "normally" for task properties we should do that here.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -53,18 +58,54 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
...
+                                    .setMessage("Gradle managed property getter methods must be abstract. "
+                                            + "Declare this method as 'public abstract', e.g., "
+                                            + "'public abstract Property<Integer> getFoo();'. "
+                                            + "This enables Gradle to inject the property implementation "
+                                            + "automatically, removing boilerplate and supporting the Groovy DSL.")
```
CRogers: Might as well extract this useful common bit of error message to a constant.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -53,18 +58,54 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
...
+                return buildDescription(tree).build();
+            }
+            return tree.getMembers().stream()
+                    .filter(member -> member instanceof MethodTree)
+                    .map(member -> (MethodTree) member)
+                    .filter(method -> isManagedPropertyGetter(method, state))
+                    .map(method -> {
+                        if (!method.getModifiers().getFlags().contains(Modifier.ABSTRACT)) {
+                            return buildDescription(method)
+                                    .setMessage("Gradle managed property getter methods must be abstract. "
```
CRogers: I would not assume people know what a "Gradle managed property" is and instead say "properties on Tasks should be declared abstract." followed by the rest of the error message.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
CRogers: We should definitely tell people about `getFoo()` enabling `foo = 3` syntax in the error message.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeTest.java
```diff
@@ -57,6 +57,67 @@ void interface_task_should_be_fine() {
...
+
+        @Test
+        void abstract_task_with_property_getter_not_starting_with_get_should_fail() {
+            test(
+                    """
+                    import org.gradle.api.DefaultTask;
+                    import org.gradle.api.provider.Property;
+                    abstract class Test extends DefaultTask {
+                        // BUG: Diagnostic contains: must be named starting with 'get'
+                        public abstract Property<String> name();
```
CRogers, line 108: There are valid reasons for doing this - but I think it's still best to error here. People can suppress the check for real instances where this is needed for backcompat.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeTest.java
```diff
@@ -57,6 +57,67 @@ void interface_task_should_be_fine() {
...
+
+        @Test
+        void abstract_task_with_non_abstract_property_getter_should_fail() {
+            test(
+                    """
+                    import org.gradle.api.DefaultTask;
+                    import org.gradle.api.provider.Property;
+                    abstract class Test extends DefaultTask {
+                        // BUG: Diagnostic contains: must be abstract
+                        public Property<String> getFoo() { return null; }
```
CRogers: These are definitely still valid cases to check, but the one we really want to stop people doing is this style:

```java
class SomeTask extends DefaultTask {
    private final Property<String> property;

    public SomeTask() {
         this.property = getProject().getObjectFactory().property(String.class);
    }
```

and make an error on the field. Maybe follow up in a later PR?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -53,18 +58,54 @@ public final class NonAbstractGradleType extends GradleGuideBugChecker
...
 
+    private static final Set<String> SUPPORTED_PROPERTY_TYPES = Set.of(
+            "org.gradle.api.provider.Property",
+            "org.gradle.api.provider.ListProperty",
+            "org.gradle.api.provider.SetProperty",
+            "org.gradle.api.provider.MapProperty",
+            "org.gradle.api.file.RegularFileProperty",
+            "org.gradle.api.file.DirectoryProperty",
+            "org.gradle.api.provider.Provider",
+            "org.gradle.api.model.ObjectFactory");
```
CRogers: Not sure how I feel about `ObjectFactory` - feels different to the others. What case were we thinking of this erroring on?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -88,9 +129,45 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
+        }
+
+        ClassSymbol extSym = (ClassSymbol) extensionType.tsym;
+        // Stream over the symbol table for methods
+        return StreamSupport.stream(extSym.members().getSymbols().spliterator(), false)
+                .filter(memberSym -> memberSym instanceof MethodSymbol)
+                .map(memberSym -> (MethodSymbol) memberSym)
+                .filter(memberSym -> isManagedPropertyGetter(memberSym, state))
+                .map(memberSym -> {
+                    if ((memberSym.flags() & Flags.ABSTRACT) == 0) {
```
CRogers: I think you can generify `isAbstract` below to take any `Symbol` and use the `getModifiers` method to see if it contains `ABSTRACT`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
FinlayRJW: good to know thanks! I've moved the detection for tasks into `matchMethod`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
FinlayRJW: removed

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeTest.java
FinlayRJW: added this as it was not much of an addition but can split into another PR if that is easier

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -70,6 +106,55 @@ public Description matchClass(ClassTree tree, VisitorState state) {
...
+        if (!isAbstract(ASTHelpers.getSymbol(method))) {
+            return buildDescription(method)
+                    .setMessage(ABSTRACT_PROPERTY_METHOD_MSG)
+                    .build();
+        }
+        if (!method.getName().toString().startsWith("get")) {
+            return buildDescription(method)
+                    .setMessage(PROPERTY_GETTER_NAMING_MSG)
+                    .build();
+        }
```
CRogers, line 129: Might as well report both at once - the way to do this is give the description to `state.reportMatch` rather than returning, so you can report multiple issues.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -46,20 +53,49 @@
...
+            + "the Groovy DSL (e.g. `foo = 3`).";
+
+    private static final Set<String> SUPPORTED_PROPERTY_TYPES = Set.of(
+            "org.gradle.api.provider.Property",
+            "org.gradle.api.provider.ListProperty",
+            "org.gradle.api.provider.SetProperty",
+            "org.gradle.api.provider.MapProperty",
+            "org.gradle.api.file.RegularFileProperty",
+            "org.gradle.api.file.DirectoryProperty",
+            "org.gradle.api.provider.Provider");
```
CRogers: I think all of these can be replaced by doing a `isSubType` check on `Provider`.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
CRogers: You can find subtypes by doing `state.getTypes().isSubType()`.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
```diff
@@ -46,20 +53,49 @@
...
+    private static final String PROPERTY_GETTER_NAMING_MSG =
+            "Properties on Tasks or Extensions should be named starting with 'get', e.g., 'getFoo'. This naming "
+                    + "convention is required for Gradle to recognize and manage the property, enabling automatic "
+                    + "property wiring and Groovy DSL support, this will automatically this will make the `foo = 3` "
+                    + "groovy syntax work of the box.";
+
+    private static final String PROPERTY_FIELD_MSG = "Do not declare Property fields directly on Tasks or Extensions. "
+            + "Instead, declare an abstract getter method, e.g., 'public abstract Property<String> getFoo();'. This "
+            + "enables Gradle to inject the property implementation automatically, removes boilerplate, and supports "
+            + "the Groovy DSL (e.g. `foo = 3`).";
```
CRogers, line 86: Nice message

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
CRogers: Might be worth writing a check about this case just to be sure...

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleType.java
FinlayRJW: cool switched to using `ASTHelpers.isSubtype`

FinlayRJW: I've also switched a few of the tests to check other types e.g. RegularFileProperty

FinlayRJW: Hmm this doesn't seem to work for `NamedDomainObjectContainer`

FinlayRJW: I've added checks for if subtype of FileCollection or DomainObjectCollection as well as I think these should be included

PR #121, opened 2025-05-20
Errorprone to prevent the assignment of values to class fields in tasks or extensions 

FinlayRJW:
## Before this PR
No error prone to prevent the following bad gradle in tasks or extensions:

```java
abstract class Test extends DefaultTask {
    private final Property<String> foo;
    public Test() {
        this.foo = getProject().getObjects().property(String.class);
    }
}
```

## After this PR
Now this will cause an errorprone compilation error

==COMMIT_MSG==
Errorprone to prevent the assignment of values to class fields in tasks or extensions 
==COMMIT_MSG==

### DIFF
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionClassMatcher.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionClassMatcher.java
@@ -1,0 +1,62 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.guide.errorprone;
+
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.suppliers.Supplier;
+import com.google.errorprone.suppliers.Suppliers;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
+import com.sun.tools.javac.code.Type;
+import java.util.Optional;
+
+public interface ExtensionClassMatcher extends BugChecker.MethodInvocationTreeMatcher {
+
+    Matcher<ExpressionTree> IS_EXTENSION_CREATE = Matchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.plugins.ExtensionContainer")
+            .named("create");
+    Supplier<Type> CLASS_TYPE_SUPPLIER = Suppliers.typeFromString("java.lang.Class");
+
+    default Optional<ClassSymbol> matchExtensionClass(MethodInvocationTree tree, VisitorState state) {
+        // We have to do our checks on Extensions where they are created rather than on the Extension types themselves
+        // This is because Extension types do not extend a class or implement an interface. There's no way to tell if
+        // a certain class is an extension or not just looking at its type declaration alone.. Especially given
+        // there are many other types with names ending in Extension, eg Junit 5 extensions.
+        if (!IS_EXTENSION_CREATE.matches(tree, state)) {
+            return Optional.empty();
+        }
+
+        // We need at least 2 arguments (name and class)
+        if (tree.getArguments().size() < 2) {
+            return Optional.empty();
+        }
+
+        // Get the second argument which should be the class type
+        ExpressionTree classArg = tree.getArguments().get(1);
+        return Optional.ofNullable(ASTHelpers.getType(classArg))
+                .filter(argType -> ASTHelpers.isSubtype(argType, CLASS_TYPE_SUPPLIER.get(state), state))
+                .filter(argType -> !argType.getTypeArguments().isEmpty())
+                .map(argType -> argType.getTypeArguments().get(0))
+                .filter(extensionType -> extensionType.tsym instanceof ClassSymbol)
+                .map(extensionType -> (ClassSymbol) extensionType.tsym);
+    }
+}
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionClassMatcher.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionClassMatcher.java
@@ -17,7 +17,8 @@
 package com.palantir.gradle.guide.errorprone;
 
 import com.google.errorprone.VisitorState;
-import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
+import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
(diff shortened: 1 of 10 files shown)


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeFields.java
```diff
@@ -0,0 +1,160 @@
...
+        // Get the second argument which should be the class type
+        ExpressionTree classArg = tree.getArguments().get(1);
+
+        Optional<Type> extensionTypeOpt = typeArgumentFromPossibleClassType(state, classArg);
+        if (extensionTypeOpt.isEmpty()) {
+            return Description.NO_MATCH;
+        }
+        Type extensionType = extensionTypeOpt.get();
+
+        ClassSymbol extSym = (ClassSymbol) extensionType.tsym;
```
CRogers: Shall we extract out the common interface here? ie an interface that extends MethodTreeMatcher and has a default method that both checks implement? Then we can just match on this classSymbol with some `matchExtensionClass` that the checks implement?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeFields.java
```diff
@@ -0,0 +1,160 @@
...
+            return Description.NO_MATCH;
+        }
+
+        if (!IS_TASK.matches(enclosingClass, state)) {
+            return Description.NO_MATCH;
+        }
+
+        if (!enclosingClass.getMembers().contains(tree)) {
+            return Description.NO_MATCH;
+        }
```
CRogers: I think for the enclosingcClass tree/getMembers, instead you can just check if `state.getPath().getParentPath()` is a ClassTree?

FinlayRJW: yep, looks like the magic combination was `state.getPath().getParentPath().getLeaf()`

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeFieldsTest.java
```diff
@@ -0,0 +1,398 @@
...
+                    .addSourceLines(
+                            "FooExtension.java",
+                            "import org.gradle.api.provider.Property;",
+                            "import org.gradle.api.Project;",
+                            "abstract class FooExtension {",
+                            "    private final Property<String> property;",
+                            "    public FooExtension(Project project) {",
+                            "        this.property = project.getObjects().property(String.class);",
+                            "    }",
+                            "}")
```
CRogers: Just make this a text block

CRogers: Same for all the others

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeFieldsTest.java
```diff
@@ -0,0 +1,398 @@
...
+                            import org.gradle.api.Plugin;
+                            import org.gradle.api.Project;
+                            public class FooPlugin implements Plugin<Project> {
+                                @Override
+                                public void apply(Project project) {
+                                    // BUG: Diagnostic contains: declare class fields directly on Tasks or Extensions
+                                    project.getExtensions().create("foo", FooExtension.class);
+                                }
+                            }
+                            """)
```
CRogers: Extract this out to a variable to avoid duplication

CRogers: Even better, make a new method to run this test inside this nested test class, you can then add `@Language("Java")` to get syntax highlighting on the extension code.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeFields.java
FinlayRJW: so I have added a class `ExtensionClassMatcher` which extends `BugChecker.MethodInvocationTreeMatcher` and adds a default `matchExtensionClass` method which the other classes can then use e.g:

```java
    @Override
    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
        return matchExtensionClass(tree, state)
                .filter(extSym -> {
                    Type type = extSym.type;
                    return !(isTypeAbstract(type) || type.isInterface());
                })
                .map(nonAbstractType -> buildDescription(tree).build())
                .orElse(Description.NO_MATCH);
    }
```

is this what you had in mind?

FinlayRJW: @CRogers  I've reworked this as discussed to below:

```java
    public Description matchExtensionClass(ClassSymbol extensionClass, MethodInvocationTree tree, VisitorState state) {
        if (isTypeAbstract(extensionClass.type) || extensionClass.type.isInterface()) {
            return Description.NO_MATCH;
        }

        return describeMatch(tree);
    }
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionClassMatcher.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Optional;
+
+public interface ExtensionClassMatcher extends MethodInvocationTreeMatcher {
+
+    Matcher<ExpressionTree> IS_EXTENSION_CREATE = Matchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.plugins.ExtensionContainer")
+            .named("create");
+    Supplier<Type> CLASS_TYPE_SUPPLIER = Suppliers.typeFromString("java.lang.Class");
+
+    Description matchExtensionClass(ClassSymbol extensionClass, MethodInvocationTree tree, VisitorState state);
```
CRogers, line 40: I'd give this `tree` a descriptive name, like `extensionContainerCreateTree` to reflect what it is.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeFields.java
```diff
@@ -0,0 +1,118 @@
...
+    private static final Matcher<ClassTree> IS_TASK = Matchers.isSubtypeOf("org.gradle.api.Task");
+    private static final Supplier<Type> PROVIDER_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.provider.Provider");
+    private static final Supplier<Type> DOMAIN_OBJECT_COLLECTION_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.DomainObjectCollection");
+    private static final Supplier<Type> FILE_COLLECTION_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.file.FileCollection");
+
+    public static final String SUMMARY =
+            "Do not declare class fields directly on Tasks or Extensions. Instead, declare an abstract getter "
```
CRogers: > Do not declare class fields directly on Tasks or Extensions

I would change to:

> Do not declare Properties, FileCollections and other Gradle managed types as fields directly on Tasks or Extensions

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/NonAbstractGradleTypeFields.java
```diff
@@ -0,0 +1,118 @@
...
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
+import com.sun.tools.javac.code.Symbol.VarSymbol;
+import com.sun.tools.javac.code.Type;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+@AutoService(BugChecker.class)
+@BugPattern(severity = SeverityLevel.ERROR, summary = NonAbstractGradleTypeFields.SUMMARY)
+public final class NonAbstractGradleTypeFields extends GradleGuideBugChecker
```
CRogers: Can we call this one `GradleTypesAsFields`?

CRogers:
:+1: given comments

FinlayRJW:
> 👍 given comments

I think it has revoked your approval after I made the changes

CRogers:
:+1:

PR #125, opened 2025-05-29
Errorprone to check that abstract methods in Tasks or Extensions that return Gradle managed types start with 'get'

FinlayRJW:
## Before this PR
Had no error prone to ensure this gradle style

## After this PR
Now if a method that is in a task or extension and is abstract does not start with get we add an error prone e.g.

```
public abstract Property<String> foo()
```

would result in an error prone

==COMMIT_MSG==
Errorprone to check that abstract methods in Tasks or Extensions that return Gradle managed types start with 'get'
==COMMIT_MSG==

## Possible downsides?
It is not always true that get should be prefixed - but in these rare situations it can be suppressed

### DIFF
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GradleManagedTypeGetPrefix.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GradleManagedTypeGetPrefix.java
@@ -29,17 +29,19 @@ import com.google.errorprone.suppliers.Suppliers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.MethodInvocationTree;
-import com.sun.source.tree.VariableTree;
+import com.sun.source.tree.MethodTree;
+import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symbol.ClassSymbol;
-import com.sun.tools.javac.code.Symbol.VarSymbol;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Type;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
+import javax.lang.model.element.Modifier;
 
 @AutoService(BugChecker.class)
-@BugPattern(severity = SeverityLevel.ERROR, summary = GradleTypesAsFields.SUMMARY)
-public final class GradleTypesAsFields extends GradleGuideBugChecker
-        implements BugChecker.VariableTreeMatcher, ExtensionClassMatcher {
+@BugPattern(severity = SeverityLevel.ERROR, summary = GradleManagedTypeGetPrefix.SUMMARY)
+public final class GradleManagedTypeGetPrefix extends GradleGuideBugChecker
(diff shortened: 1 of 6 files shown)


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GradleManagedTypeGetPrefix.java
```diff
@@ -0,0 +1,139 @@
...
+    private static final Supplier<Type> PROVIDER_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.provider.Provider");
+    private static final Supplier<Type> DOMAIN_OBJECT_COLLECTION_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.DomainObjectCollection");
+    private static final Supplier<Type> FILE_COLLECTION_TYPE_SUPPLIER =
+            Suppliers.typeFromString("org.gradle.api.file.FileCollection");
+
+    public static final String SUMMARY =
+            "Abstract methods in Tasks or Extensions that return Gradle managed types should start with 'get'. "
+                    + "This naming convention ensures consistency and allows Gradle to handle property injection "
```
CRogers: I would remove the "ensures consistency" - we normal eschew using `get` prefixes and only really do it for the Gradle magic reason

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GradleManagedTypeGetPrefix.java
```diff
@@ -0,0 +1,139 @@
...
+                .filter(methodSym -> isManagedPropertyType(methodSym.getReturnType(), state))
+                .map(MethodSymbol::getSimpleName)
+                .filter(simpleName -> !simpleName.toString().startsWith("get"))
+                .map(Object::toString)
+                .collect(Collectors.joining(", "));
+
+        if (!methodNames.isEmpty()) {
+            return buildDescription(extensionContainerCreateTree)
+                    .setMessage("Within the " + extensionClass.getSimpleName()
+                            + " extension, the following methods do not start with 'get': " + methodNames + "\n"
```
CRogers, line 98: Nice

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GradleManagedTypeGetPrefix.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private static boolean isAbstract(Symbol symbol) {
+        return symbol != null && symbol.getModifiers().contains(Modifier.ABSTRACT);
+    }
+
+    private static boolean isManagedPropertyType(Type type, VisitorState state) {
+        return isSubtypeOfProvider(type, state)
+                || isSubtypeOfDomainObjectCollection(type, state)
+                || isSubtypeOfFileCollection(type, state);
+    }
```
CRogers: Don't we have this logic already in another error prone? If so, we should extract out to another class both can use.

FinlayRJW: yep - moved to a utility class `GradleManagedTypes`

CRogers (approved):

PR #127, opened 2025-05-30
Ensure that extensions are only added via `.create`

FinlayRJW:
## Before this PR
had no errorprone to prevent behaviour like:

```java
project.getExtensions().add("foo", new fooExtension());
```

## After this PR
We now have an errorprone with autofix to prevent users from doing this it will auto fix as below:

```java
project.getExtensions().add("foo", new FooExtension("a", "b"));
```
-->
```java
project.getExtensions().create("foo", FooExtension.class, "a", "b");
```

==COMMIT_MSG==
Ensure that extensions are only added via `.create`
==COMMIT_MSG==

### DIFF
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
@@ -33,8 +33,10 @@ import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
 import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
 import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.util.Name;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -44,9 +46,8 @@ import java.util.stream.Stream;
 @AutoService(BugChecker.class)
 @BugPattern(
         summary = "Gradle extensions should be registered using `create` rather than by constructing and passing a new "
-                + "instance to `add`. `create` ensures that Gradle manages the extension's lifecycle, type, "
-                + "and dependency injection correctly, while `add` with a manually constructed instance can "
-                + "lead to misconfigured extensions and bypasses Gradle's internal management.",
+                + "instance to `add`. Using `create` enables Gradle to properly manage the extension's type. "
+                + "Manual construction with `add` can interfere with certain error-prone checks.",
         severity = SeverityLevel.ERROR)
 public final class ExtensionsUseCreate extends GradleGuideBugChecker implements BugChecker.MethodInvocationTreeMatcher {
 
@@ -68,24 +69,25 @@ public final class ExtensionsUseCreate extends GradleGuideBugChecker implements 
         ExpressionTree instanceArg = args.get(args.size() - 1);
 
         if (instanceArg instanceof MethodInvocationTree methodArg) {
-            if (isIgnoredMethodInvocation(methodArg, state)) {
+            if (isIgnoredMethodInvocation(methodArg)) {
                 return Description.NO_MATCH;
             }
         }
 
         if (instanceArg instanceof NewClassTree newClass) {
-            return buildDescription(tree)
-                    .addFix(suggestCreateFix(tree, newClass, Optional.empty(), state)
-                            .build())
-                    .build();
+            Optional<SuggestedFixBuilder> maybeFix = suggestCreateFix(tree, newClass, Optional.empty(), state);
+            Description.Builder description = buildDescription(tree);
+            maybeFix.ifPresent(fix -> description.addFix(fix.build()));
+            return description.build();
         }
 
         if (instanceArg instanceof IdentifierTree ident) {
             Optional<VariableTree> varDecl = findVariableDeclaration(ident, state);
             if (varDecl.isPresent() && varDecl.get().getInitializer() instanceof NewClassTree newClass) {
-                return buildDescription(tree)
-                        .addFix(suggestCreateFix(tree, newClass, varDecl, state).build())
-                        .build();
+                Optional<SuggestedFixBuilder> maybeFix = suggestCreateFix(tree, newClass, varDecl, state);
+                Description.Builder description = buildDescription(tree);
+                maybeFix.ifPresent(fix -> description.addFix(fix.build()));
+                return description.build();
             }
         }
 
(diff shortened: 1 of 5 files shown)


guide/managed-types-and-properties.md
```diff
@@ -60,6 +60,25 @@ This is very similar to [the commonly used `immutables` library](https://immutab
 
 It is possible to write extensions and tasks directly as in the `HelloExtension_Generated` class above. If you look into the depths of older Palantir Gradle code you will find many instances of this, but this is no longer recommended. All new code should use Managed Types.
 
+### Why prefer `create` over `add` for extensions?
```
FinlayRJW, line 63: Not sure if there is where this should live but only instance of using `project.getExtensions().create(...)` in the guide

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
```diff
@@ -0,0 +1,154 @@
...
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.joining(", ", extensionsReceiver + ".create(", ")"));
+
+        SuggestedFixBuilder fix = new SuggestedFixBuilder();
+        fix.replace(addCall, createCall);
+        maybeVarDecl.ifPresent(fix::delete);
+        return fix;
+    }
+
+    private boolean isIgnoredMethodInvocation(MethodInvocationTree methodArg, VisitorState state) {
```
FinlayRJW: thought we might as well check if it is getting an existing extension can remove if we think this might be brittle

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreateTest.java
```diff
@@ -0,0 +1,375 @@
...
+ */
+package com.palantir.gradle.guide.errorprone;
+
+import com.google.errorprone.CompilationTestHelper;
+import com.palantir.gradle.guide.helpers.RefactoringValidator;
+import org.intellij.lang.annotations.Language;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class ExtensionsUseCreateTest {
```
CRogers, line 25: These tests look great, very thorough!

guide/managed-types-and-properties.md
```diff
@@ -60,6 +60,25 @@ This is very similar to [the commonly used `immutables` library](https://immutab
...
+```
+rather than
+
+```java
+project.getExtensions().add("name", new MyExtension(/* … */));
+```
+create asks Gradle to
+- build a managed instance (using Gradle’s code-generation),
+- wire that instance into Gradle’s lifecycle and dependency-injection system
+- keep all properties lazily realizable so they can still be configured after creation but before task-graph realisation.
```
CRogers: Not sure about this one

CRogers: I don't think there's any difference between when properties are realised between the two.

guide/managed-types-and-properties.md
```diff
@@ -60,6 +60,25 @@ This is very similar to [the commonly used `immutables` library](https://immutab
...
+
+```java
+project.getExtensions().add("name", new MyExtension(/* … */));
+```
+create asks Gradle to
+- build a managed instance (using Gradle’s code-generation),
+- wire that instance into Gradle’s lifecycle and dependency-injection system
+- keep all properties lazily realizable so they can still be configured after creation but before task-graph realisation. 
+
+Passing a manually constructed new MyExtension(…) to add skips all of this—the object is unmanaged, Gradle cannot inject services, and property realisation becomes eager, often leading to brittle or misconfigured builds.
```
CRogers: Ditto about realisation here.

guide/managed-types-and-properties.md
CRogers: Best place for it for now

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        summary = "Gradle extensions should be registered using `create` rather than by constructing and passing a new "
+                + "instance to `add`. `create` ensures that Gradle manages the extension's lifecycle, type, "
+                + "and dependency injection correctly, while `add` with a manually constructed instance can "
+                + "lead to misconfigured extensions and bypasses Gradle's internal management.",
```
CRogers: > while `add` with a manually constructed instance can lead to misconfigured extensions and bypasses Gradle's internal management.

I think this bit is a bit strong. I think even with `.add` it should probably be fine; I mainly want to get rid of this kind of usage as 1) You can't do the abstract pattern (without doing `ObjectFactory#newInstance`) and 2) it stops the our other error prones working.

guide/managed-types-and-properties.md
CRogers: >often leading to brittle or misconfigured builds.

this is probably also a bit strong

guide/managed-types-and-properties.md
```diff
@@ -60,6 +60,25 @@ This is very similar to [the commonly used `immutables` library](https://immutab
 
 It is possible to write extensions and tasks directly as in the `HelloExtension_Generated` class above. If you look into the depths of older Palantir Gradle code you will find many instances of this, but this is no longer recommended. All new code should use Managed Types.
 
+### Why prefer `create` over `add` for extensions?
+
+Always register an extension via
```
CRogers: I would go always -> prefer

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
```diff
@@ -0,0 +1,154 @@
...
+        fix.replace(addCall, createCall);
+        maybeVarDecl.ifPresent(fix::delete);
+        return fix;
+    }
+
+    private boolean isIgnoredMethodInvocation(MethodInvocationTree methodArg, VisitorState state) {
+        String calledName = state.getSourceForNode(methodArg.getMethodSelect());
+        return calledName == null
+                || calledName.endsWith(".getByName")
+                || calledName.endsWith(".findByName")
```
CRogers: Might as well throw `getByType` in there too.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
```diff
@@ -0,0 +1,154 @@
...
+                .collect(Collectors.joining(", ", extensionsReceiver + ".create(", ")"));
+
+        SuggestedFixBuilder fix = new SuggestedFixBuilder();
+        fix.replace(addCall, createCall);
+        maybeVarDecl.ifPresent(fix::delete);
+        return fix;
+    }
+
+    private boolean isIgnoredMethodInvocation(MethodInvocationTree methodArg, VisitorState state) {
+        String calledName = state.getSourceForNode(methodArg.getMethodSelect());
```
CRogers: Rather than use source code, I think you can do `ASTHelpers.getSymbol(methodArg)` then maybe `.getSimpleName()` on the `MethodSymbol`?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private static SuggestedFixBuilder suggestCreateFix(
+            MethodInvocationTree addCall,
+            NewClassTree newClass,
+            Optional<VariableTree> maybeVarDecl,
+            VisitorState state) {
+
+        String extensionsReceiver = Optional.ofNullable(state.getSourceForNode(addCall.getMethodSelect()))
+                .orElse("/* project.getExtensions() */")
```
CRogers: What are these commented bits for?

FinlayRJW: if it can't find the node it will put the comment in as part of the auto fix e.g. `/* project.getExtensions() */.create(...)`

FinlayRJW: we should probably just not suggest a fix if we fail though - will update

guide/managed-types-and-properties.md
FinlayRJW: replaced this line with:

This is mostly stylistic but is important as our Errorprones are built to detect extensions instantiated via `create`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreate.java
FinlayRJW: updated to:

```java
        summary = "Gradle extensions should be registered using `create` rather than by constructing and passing a new "
                + "instance to `add`. Using `create` enables Gradle to properly manage the extension's type. "
                + "Manual construction with `add` can interfere with certain error-prone checks.",
```

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/ExtensionsUseCreateTest.java
```diff
@@ -0,0 +1,375 @@
...
+                    .addOutputLines("Test.java", expected)
+                    .doTest();
+        }
+
+        private void refactoringNoop(@Language("Java") String code) {
+            bestEffortRefactoringValidator()
+                    .addInputLines("Test.java", code)
+                    .expectUnchanged()
+                    .doTest();
+        }
```
kelvinou01, line 380: Hmm let's not use the `bestEffortRefactoringValidator` if this error-prone doesn't use `BestEffortMode`!

kelvinou01 (commented):
Overall LGTM! Just a minor nit then we can ship this.

kelvinou01 (approved):

PR #142, opened 2025-06-23
Add the `IllegalMethodCalledDuringTaskExecution` ErrorProne to prevent `getProject()` calls from within a task action

kelvinou01:
## Before this PR

This errorprone prevents the one of the most common violations of [configuration cache requirements](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements) — invoking `getProject()` within a `@TaskAction` or `Action<Task>::execute`. It also prevents regressions after we've rolled out the plugin.

## After this PR

==COMMIT_MSG==
Add the `IllegalMethodCalledDuringTaskExecution` ErrorProne to prevent `getProject()` calls from within a task action
==COMMIT_MSG==

## Possible downsides?

This PR does not catch the case below. It will be handled in a following PR.

```java
import org.gradle.api.DefaultTask;
import org.gradle.api.Plugin;
import org.gradle.api.Project;
import org.gradle.api.tasks.TaskAction;

public abstract class PrintProjectName extends DefaultTask {
    private boolean getProjectName(){
        return getProject().getName();
    }
    
    @TaskAction
    public final void action() {
        System.out.println("Project name: " + getProjectName());
    }
}
```

FLUPs: https://github.com/palantir/gradle-guide/pull/160, https://github.com/palantir/gradle-guide/pull/173


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public Description matchMethod(MethodTree tree, VisitorState state) {
+        boolean isTaskAction = tree.getModifiers().getAnnotations().stream()
+                .anyMatch(annotation ->
+                        state.getSourceForNode(annotation.getAnnotationType()).contains("TaskAction"));
+        if (!isTaskAction) {
+            return Description.NO_MATCH;
+        }
+
+        if (checkForMethodCall(tree.getBody(), state, TASK_GET_PROJECT_METHOD)) {
```
kelvinou01: As discussed in person, let's

1. specify the line which calls getProject(), rather than the enclosing task action method
3. return an error for all offending lines, so our configuration cache problems metrics picks up on these violations

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,88 @@
...
+            return buildDescription(tree)
+                    .setMessage("Don't call getProject() in task actions")
+                    .build();
+        } else {
+            return Description.NO_MATCH;
+        }
+    }
+
+    private boolean checkForMethodCall(Tree tree, VisitorState state, Matcher<ExpressionTree> matcher) {
+        return new TreeScanner<Boolean, Void>() {
```
kelvinou01: As discussed in person, Errorprone already scans the AST, so another AST scan within *an* Errorprone could be a source of a O(n^2) nightmare. However, in this case it should be fine, as we only walk down TaskActions.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/GetProjectInvocationsTest.java
```diff
@@ -0,0 +1,168 @@
...
+                    // BUG: Diagnostic contains: Don't call `getProject()` in task actions
+                    String projectName = getProject().getName();
+                    System.out.println("Project name: " + projectName);
+
+                    // BUG: Diagnostic contains: Don't call `getProject()` in task actions
+                    String projectPath = getProject().getPath();
+                    System.out.println("Project path: " + projectPath);
+                }
+            }
+            """;
```
CRogers: I would just inline these as they're only used once. 

Also, you can extract the `compilationTestHelper.addSourceLines("CustomTask.java", source).doTest()` to a method with signature `test(@Language("Java") String source)` you do not need the `// language=Java` each time. [Example](https://sourcegraph.palantir.build/github.com/palantir/gradle-guide@6b7ecd142331ff65acf410236263ebb1a767dcd9/-/blob/gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/GradleManagedTypeGetPrefixTest.java?L235:32-235:49)

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+    private static final Supplier<ClassSymbol> TASK_SYM =
+            VisitorState.memoize(s -> (ClassSymbol) s.getSymbolFromString("org.gradle.api.Task"));
+
+    private static final Matcher<ExpressionTree> TASK_GET_PROJECT_METHOD = MethodMatchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.Task")
+            .named("getProject");
+
+    public static final String VIOLATION_MESSAGE = "Don't call `getProject()` in task actions";
+
+    public static final String SUMMARY =
```
CRogers: I would just inline this in the `@BugPattern`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+            """
+        Don't call `getProject()` in task actions. Large, mutable Gradle model types like `Gradle`, `Settings`, or
+        `Project` should not be passed into tasks as inputs. Instead, your tasks should take in the "smallest" type
+        required for the task's functionality. For example, instead of taking in `Project` to later do
```
CRogers: >Large, mutable Gradle model types like `Gradle`, `Settings`, or `Project` should not be passed into tasks as inputs. Instead, your tasks should take in the "smallest" type required for the task's functionality.

This is a different check imo. This check is only about calling `getProject()` within a task action, nothing to do with inputs.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+        `project.version`, you should declare the project version as a `Property<String>`.
+        Doing so improves performance in two ways:
+        1. It makes your tasks compatible with the configuration cache
+        2. It prevents tasks from being rendered out-of-date by a mutation unrelated to the task, e.g. to `project.name`
```
CRogers: For my own understanding: Is this something that Gradle does? If you call `getProject()` Gradle will regard stuff like project name as inputs to the task automatically? Tbf, I think if you rename a project, all the tasks in it get invalidated as they are now "different tasks" because their task path is different - ie gone from `:foo:task` to `:bar:task`.

CRogers: We should add another reason of "prevents tasks in the same project from running in parallel with this task, as mutable project state may be modified"

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    private static boolean isTaskAction(MethodTree tree, VisitorState state) {
+        return tree.getModifiers().getAnnotations().stream().anyMatch(annotation -> {
+            String source = state.getSourceForNode(annotation.getAnnotationType());
+            return source.equals("TaskAction");
+        });
```
CRogers: We should avoid looking at source where ever possible. It's sometimes not available for one, it's also less precise than looking at the AST.

Instead I think we can just use `Matchers.hasAnnotation("org.gradle.api.tasks.TaskAction")`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+        ClassSymbol classSym = ASTHelpers.getSymbol(tree);
+        ClassSymbol actionSym = ACTION_SYM.get(state);
+        ClassSymbol taskSym = TASK_SYM.get(state);
+
+        List<Type> interfaces = ((ClassType) classSym.type).interfaces_field;
+        return interfaces.stream()
+                .anyMatch(ifaceType -> ifaceType.tsym.equals(actionSym)
+                        && ifaceType.getTypeArguments().size() == 1
+                        && ifaceType.getTypeArguments().get(0).tsym.equals(taskSym));
+    }
```
CRogers: I think rather than doing all the symbol stuff, you could just use a matcher like:

```java
Matchers.instanceMethod()
                    .onDescendantOf("org.gradle.api.Action")
                    .named("execute")
                    .withParameters("org.gradle.api.Task");
```

and just depend on the fact that if it's an `Action<Task>` the execute method will take a `Task`, and not have to worry about checking the type parameters of the interfaces.

Additionally, I think this implementation only checks up one level of supertypes - where as `onDescendantOf` should handle a class implementing an interface that itself implements `Action`.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private void reportAllViolations(
+            Tree tree, VisitorState state, Matcher<ExpressionTree> violationMatcher, String violationMessage) {
+        new TreeScanner<Boolean, Void>() {
```
CRogers: `SuppressibleTreeScanner` or we'll never be able to auto-suppress these.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+import com.sun.source.util.TreeScanner;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
+import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.code.Type.ClassType;
+import java.util.List;
+
+@AutoService(BugChecker.class)
+@BugPattern(severity = SeverityLevel.ERROR, summary = GetProjectInvocations.SUMMARY)
+public final class GetProjectInvocations extends GradleGuideBugChecker implements BugChecker.MethodTreeMatcher {
```
CRogers: Not convinced about `GetProjectInvocations` as a name. I would call it `GetProjectDuringTaskExecution`. But there there are also other methods we care about people not calling during task execution (`getDependsOn()`, `getTaskDependencies()`, `getConvention()`, `getExtensions()`) which will likely add in a later PR. So because this is quite a bit more general (I doubt we'd make extra checks for these), maybe it should be something like `IllegalMethodCalledDuringTaskExecution`? It's a bit long, open to ideas.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,134 @@
...
+            .named("getProject");
+
+    public static final String VIOLATION_MESSAGE = "Don't call `getProject()` in task actions";
+
+    public static final String SUMMARY =
+            """
+        Don't call `getProject()` in task actions. Large, mutable Gradle model types like `Gradle`, `Settings`, or
+        `Project` should not be passed into tasks as inputs. Instead, your tasks should take in the "smallest" type
+        required for the task's functionality. For example, instead of taking in `Project` to later do
+        `project.version`, you should declare the project version as a `Property<String>`.
```
CRogers: Might want to talk about injecting services too (eg using injected `FileOperations` rather than `project#copy`, or `ExecOperations` rather than `project#exec` - basically everything [here](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution) - might want to add a page to the guide (with examples in Java) about this.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: Gotcha, I'll make the change.

> It's sometimes not available for one

When so?

> it's also less precise than looking at the AST

Seems like Matchers.hasAnnotation matches with Symbols. Is that generally what we want to do, for precision?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: I wanted to use `Matchers`, but they match on method invocations rather than method declarations. 

The matcher 

```java
Matchers.instanceMethod() .onDescendantOf("org.gradle.api.Action") .named("execute") .withParameters("org.gradle.api.Task");
```

takes in an `ExpressionTree`, which a method declaration is not.

I don't see any utilities in `Matchers` which act on `MethodTree`s : (

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/GetProjectInvocationsTest.java
kelvinou01: Gotcha, thanks for the pointers!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: How does `ConfigurationCacheViolation` sound? It's direct and points people towards learning about [what is a violation[(https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements). 

Or maybe `ExecutionPhaseViolation`?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: Yep, explicating tribal knowledge about service injection sounds like a good idea. I'll add it to the guide.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: Yep!

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,245 @@
+<!-- PreviousNext:START -->
+<table><tr>
+  <td>Previous: <a href="testing-and-running-your-plugins-locally.md">Testing and running your plugins</a></td>
+  <td align="center"><a href="../README.md#table-of-contents">Table of Contents</a></td>
+  <td align="right">Next: <a href="diagnosing-build-performance.md">Diagnosing Build Performance</a></td>
+</tr></table>
+<!-- PreviousNext:END -->
+
+# Adopting the Configuration Cache
```
CRogers, line 9: I think this needs a brief overview of how the configuration cache works, ie it saves the input/output properties of each task and loads it back later to avoid the configuration stage, it regards any external process/file read/gradle property etc as an input to configuration time that is evaluated every time it's run to see if the cache is busted or not.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,245 @@
...
+2. Find the configuration cache problems in the output. If there are no problems, congratulations! Your build is Configuration Cache compatible.
+3. If not, fix the problems. 
+4. Run `./gradlew build --configuration-cache` again. In our experience, this uncovers more problems that weren't there in (1). 
+
+The problems you find in step (2) are usually one of two kinds:
+
+1. **external process started `/usr/bin/git --version`**
+2. **cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, a subtype of `org.gradle.api.Project`, as these are not supported with the configuration cache**
+
+The first error occurs when you start external processes without using `ExecOperations`. Gradle wants to know about external processes
```
CRogers: >Gradle wants to know about external processes

... so it can decide whether or not it needs to rerun the configuration task as an input to the configuration cache has changed?

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,245 @@
...
+
+    @Input
+    protected abstract Property<Path> getOutputFile();
+
+    @Inject
+    protected abstract ObjectFactory getObjectFactory();
+
+    @TaskAction
+    public void compress() {
+        ZstdCompressor compressor = getObjectFactory().newInstance(ZstdCompressor.class, 0);
```
CRogers: The ideal would be to make this an `@Internal @Nested` but I wonder whether that's too many ideas to introduce here...

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: Yep, I'll update the summary to be specific to this errorprone.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: I'm not sure tbh. Let's update this section to just say things we're sure are true.

kelvinou01: Added!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: Yep! Let's use `SuppressibleTreePathScanner` for all our tree-scanning needs in gradle-guide.

guide/adopting-the-configuration-cache.md
kelvinou01: We could, but I wanted to give the example of a regular stateful java class with non-static members and constructors with arguments. I think if people know how to deal with this case, then other simpler cases should be self-evident.

But yep @Nested is a neat trick people might appreciate. I've added a small snippet explaining the use of @Nested in the guide, let me know what you think of it.

guide/adopting-the-configuration-cache.md
kelvinou01: I've added a sentence explaining this, let me know if it captures what you want to express here.

guide/adopting-the-configuration-cache.md
kelvinou01: I didn't want to repeat any information from the user guide. Do you think we should explain the fundamentals of the CC here as well? Or maybe we can point people towards that section of the guide?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
CRogers: I think `ConfigurationCacheViolation` is a bit overbroad. There are other configuration cache issues (like taking an `@Input Property<Configuration>`) which this check is not going to apply to. I think `ExecutionPhaseViolation` is also a bit vague. Let's go with `IllegalMethodCalledDuringTaskExecution`? I feel it's long but at least it's accurate.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/GetProjectInvocationsTest.java
```diff
@@ -0,0 +1,154 @@
...
+    private final CompilationTestHelper compilationTestHelper =
+            CompilationTestHelper.newInstance(GetProjectInvocations.class, getClass());
+
+    /**
+     * Tests for {@code @TaskAction}.
+     */
+    @Nested
+    class Tasks {
+        @Test
+        void getProjectWithinTaskActionsShouldFail() {
```
CRogers: We generall prefer to write test methods as English sentences with underscores. ie `getProject_within_task_actions_should_fail`. It just makes the tests a bit more readable and makes it more obvious which ones are tests.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,325 @@
...
+    public void compress() {
+        Startup startup = getObjectFactory().newInstance(Startup.class, 1000000000000000);
+        UploadUtils.upload(startup, getUploadUri());
+    }
+}
+```
+
+This way, you "propagate" the reach of the Gradle managed type system, removing boilerplate and making Configuration Cache adoption easy. 
+
+### You can inject services into Gradle-managed types
```
kelvinou01: As discussed in person, lets put this in front of "### You can make any class into a Gradle-managed type"

guide/adopting-the-configuration-cache.md
kelvinou01: As discussed in person, this guide should be relatively self-contained, and help people to understand enough to dive into the gradle user guide themselves. 

Let's explain the fundamentals of the Configuration Cache here.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = SeverityLevel.ERROR,
+        summary =
+                """
+        Don't call `getProject()` in task actions. Instead, your tasks should take in the "smallest" type
+        required for the task's functionality. For example, instead of taking in `Project` to later do
+        `project.version`, you should declare the project version as a `Property<String>`.
```
CRogers: > `@Input public abstract Property<String>`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GetProjectInvocations.java
kelvinou01: Yep, let's go with `IllegalMethodCalledDuringTaskExecution` then.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/GetProjectInvocationsTest.java
kelvinou01: Gotcha.

kelvinou01:
### Feedback from Gradle devs

> Hey, I just went through the guide and it is quite straightforward to follow, I really liked it that there was an example of fixing issues from a sample Gradle task. I just noticed one thing that I had no idea what it means: ZstdCompressor is a POJO.
Also, one suggestion I would make is that the tip on incremental configuration caching is quite below, and a bit easy to miss although it is quite useful. If you want you can move it up below the "Solving Configuration Cache problems" title

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,516 @@
...
+3. If not, fix the problems.
+4. Run `./gradlew build --configuration-cache` again. In our experience, this reveals additional issues not seen in the first run.
+
+The problems typically fall into three categories:
+
+1. **external process started `/usr/bin/git --version`**
+2. **cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, a subtype of `org.gradle.api.Project`, as these are not supported with the configuration cache**
+3. **invocation of `Task.project` at execution time is unsupported**
+
+The first error occurs when you start external processes during the configuration phase. Generally, , it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE.
```
CRogers: > , ,

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,516 @@
...
+
+The problems typically fall into three categories:
+
+1. **external process started `/usr/bin/git --version`**
+2. **cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, a subtype of `org.gradle.api.Project`, as these are not supported with the configuration cache**
+3. **invocation of `Task.project` at execution time is unsupported**
+
+The first error occurs when you start external processes during the configuration phase. Generally, , it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE. 
+
+without using `ExecOperations`. Gradle needs to be aware of all external processes so it can determine if any changes require the configuration cache to be re-run.
```
CRogers: > without using `ExecOperations`

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,516 @@
...
+
+The second and third errors occur because Gradle tasks shouldn't take mutable types as input — Gradle, Settings, Project, SourceSet, or Configuration. Having these as inputs limits task concurrency (what if two tasks mutate a `Project` concurrently?) and prevents Gradle from serializing task inputs into the cache (they are too large to be serialized).
```
CRogers: Generally, I'd wrap code references like Gradle, Settings etc in \` so they appear like `Gradle`, `Settings` etc

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,516 @@
...
+
+The first error occurs when you start external processes during the configuration phase. Generally, , it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE. 
+
+without using `ExecOperations`. Gradle needs to be aware of all external processes so it can determine if any changes require the configuration cache to be re-run.
+
+The second and third errors occur because Gradle tasks shouldn't take mutable types as input — Gradle, Settings, Project, SourceSet, or Configuration. Having these as inputs limits task concurrency (what if two tasks mutate a `Project` concurrently?) and prevents Gradle from serializing task inputs into the cache (they are too large to be serialized). 
+
+To solve the second and third errors, you can
+- Declare the smallest "surface area" your task needs as task input, e.g. take in `Property<String>` for project version, instead of taking in `Project` and accessing `project.version`
+- [Inject a service](https://docs.gradle.org/current/userguide/service_injection.html) which provides the information/operation you need
```
CRogers: I would give `FileOperations` and `ExecOperations` as examples.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,516 @@
...
+> [!TIP]
+> If you have a huge Gradle projects with many tasks, you can adopt the Configuration Cache incrementally. First, resolve all configuration phase issues, as incremental adoption only works for the execution phase. Then, apply the [gradle-incremental-configuration-cache](https://github.com/palantir/gradle-incremental-configuration-cache) plugin. Gradually add tasks to the allow list as they become compatible.
+
+
+### Before:
+
+- `CompanyReportPlugin` and `GenerateReportTask` are already [Gradle managed types](managed-types-and-properties.md)
+- `ReportGenerator` is a plain old Java object.
+- `CompanyReportPlugin::apply` starts an external process in the configuration phase, causing the configuration cache to fail
+- Furthermore, `GenerateReportTask` is calling `getProject()` at build time, causing another configuration cache failure: **invocation of `Task.project` at execution time is unsupported**
```
CRogers: No need for "Furthermore" in bullet points.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,516 @@
...
+      } catch (Exception e) {
+         throw new RuntimeException(e);
+      }
+   }
+}
+
+abstract class GenerateReportTask extends DefaultTask {
+   @TaskAction
+   public void upload() {
+      SassyReporter generator = new ReportGenerator(1000);
```
CRogers: What is `SassyReporter`? Should it be just `ReportGenerator`?

guide/adopting-the-configuration-cache.md
kelvinou01: Yep I missed these!

guide/adopting-the-configuration-cache.md
kelvinou01: Yep I've updated this section to have two examples, a FileSystemOperations one and an ExecOperations one

guide/adopting-the-configuration-cache.md
kelvinou01: That was the original name hahaha. Let's remove it

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.List;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = SeverityLevel.ERROR,
+        summary =
+                """
+        Don't call `getProject()` in task actions. Instead, your tasks should take in the "smallest" type
+        required for the task's functionality. For example, instead getProject().version, you should declare the
```
felixdesouza:
```suggestion
        required for the task's functionality. For example, instead getProject().version(), you should declare the
```

kelvinou01:
Useful tip for editing these markdown files is to turn on soft-wrap on Intellij

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+Gradle builds are split into two phases — the configuration phase creates the task graph, while the execution phase runs it. The configuration phase takes various inputs and produces a complete task graph that defines what work needs to be done and in what order.
+
+### Configuration inputs
+
+1. Gradle environment
+   - `GRADLE_USER_HOME`
+   - Gradle Daemon JVM
+2. Init scripts
+3. buildSrc and included build logic build contents (build scripts, sources, and intermediate build outputs)
```
felixdesouza:
```suggestion
3. `buildSrc` and included build logic build contents (build scripts, sources, and intermediate build outputs)
```

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+4. Build and Settings scripts, including included scripts (apply from: foo.gradle)
```
felixdesouza:
```suggestion
4. Build and Settings scripts, including included scripts (`apply from: 'foo.gradle'`)
```

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+5. Gradle configuration files (Version Catalogs, dependency verification files, dependency lock files, gradle.properties files)
```
felixdesouza:
```suggestion
5. Gradle configuration files (Version Catalogs, dependency verification files, dependency lock files, `gradle.properties` files)
```

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+6. Contents of files read at configuration time
+7. File system state checked at configuration time (file presence, directory contents, etc.)
+8. Custom ValueSource values obtained at configuration time (this also includes built-in providers, like providers.exec and providers.fileContents).
```
felixdesouza:
```suggestion
8. Custom `ValueSource` values obtained at configuration time (this also includes built-in providers, like `ProviderFactory#exec` and `ProviderFactory#fileContents`.
```

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+9. System properties used during the configuration phase
+10. Environment variables used during the configuration phase
+
+
+### Configuration outputs
+The configuration phase output is a task [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph), describing the target task and it's dependencies. Different tasks have different dependencies, and thus produce different task graphs. However, the target is always at the root of the DAG. Here are some of the task graphs of this repository:
```
felixdesouza:
```suggestion
The configuration phase output is a [directed acyclic graph (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph) of tasks, describing the target task and its dependencies. Different tasks have different dependencies, and thus produce different task graphs. However, the target task is always at the root of the DAG. Here are some of the task graphs of this repository:  
```

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+<!-- PreviousNext:END -->
+
+# Adopting the Configuration Cache
+
+The Gradle Configuration Cache is a powerful new feature. It reduces build latency by caching the task graph from the configuration phase.
+
+Builds run with the Configuration Cache optionally in Gradle 8, by default in Gradle 9, and necessarily by Gradle 10. Furthermore, your build has to meet [strict requirements](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements) to be compatible. This guide provides practical steps to adopt the Configuration Cache.
+
+
+## How does Configuration Caching work?
```
felixdesouza:
```suggestion
## How does the Configuration Cache work?
```

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+   |    +--- ...
+   +--- :spotlessCheck                                       (org.gradle.api.DefaultTask)
+   |    `--- ...
+   +--- :test                                                (org.gradle.api.tasks.testing.Test)
+   |    `--- ...
+   ...
+```
+
+When you first run a gradle task, the task graph is serialized and stored on disk. Upon a subsequent run of the same task, If none of these inputs have changed, Gradle skips the configuration phase entirely, loads the task graph from disk, and goes straight to the execution phase.
+
```
felixdesouza: You probably want to move this section slightly above `Configuration Inputs`, the main motivation/reasoning is hidden underneath walls of code.

I also think you're underselling it a bit here. 

You want something along the lines of:
* running the configuration phase can be quite costly
* assuming it is deterministic given the inputs, the output (task graph) can be cached
* Rerun all the inputs again and if so, reuse without having to recalculate inputs.

I think you've got most of it already, just more emphasis on the configuration phase being costly (in and of itself).

felixdesouza: There's also something about the verbiage which would be good to disambiguate.

It's not *just* when a task is run. It's more about the tasks requested i.e. what did you pass to `./gradlew <tasks>`. 
* It can be multiple tasks, not necessarily singular. We've been targeting the `classes` because that's the workflow we initially care about
* It can be a bit confusing to mention a task that's not in the requested task but is a dependant of one of the requested task. Especially when the log output is typically from the tasks requested.

Don't think you necessarily need to explain the distinction, but it would be worth being precise about it. e.g. mentioning what tasks requested means e.g. `./gradlew <tasks-requested...>` and talking about when you request the same set of tasks etc.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+The configuration phase typically runs faster than execution since heavy work belongs in the latter. However, without the cache, configuration phase is rerun with every single Gradle run. If you run a unit test multiple times, configuration is repeated, reproducing the same task graph every time. When the task itself is light, configuration can take up a large fraction of the latency. Configuration caching solves this by storing and reusing configuration results between runs, eliminating redundant work and speeding up iteration cycles.
```
felixdesouza: Feel like you're underselling it a bit here as well. Why is unit testing *such* a good example? because it requires a **tight feedback loop**. The task can be light, or medium, but it can still feel like it takes a while before anything "happens" i.e. paying for configuration.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+
+
+## Finding Configuration Cache problems
+
+To find Configuration Cache problems:
+
+1. Run `./gradlew build --configuration-cache`
```
felixdesouza: I think `build` is way too large a scope, and probably not that useful as a starting point. It will likely be quite overwhelming and you probably won't see results for a long time!

It's probably good as a way to maybe capture everything as lots of tasks hook on to the `build` lifecycle task transitively, but it's not a given.

I think you probably want to mention starting off with a workflow you want to "speed up" e.g. compilation workflow typically invokes: `classes` and `testClasses`, so that's what you try to find problems with.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+3. **invocation of `Task.project` at execution time is unsupported**
+
+The first error occurs when you start external processes during the configuration phase. Generally, it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE. 
+
+The second and third errors occur because Gradle tasks shouldn't take mutable types as input — `Gradle`, `Settings`, `Project`, `SourceSet`, or `Configuration`. Having these as inputs limits task concurrency (what if two tasks mutate a `Project` concurrently?) and prevents Gradle from serializing task inputs into the cache (they are too large to be serialized). 
+
+To solve the second and third errors, you can
+- Declare the smallest "surface area" your task needs as task input, e.g. take in `Property<String>` for project version, instead of taking in `Project` and accessing `project.version`
+- [Inject a service](https://docs.gradle.org/current/userguide/service_injection.html) which provides the information/operation you need
```
felixdesouza: Let's just make these into subheadings:

### Most common categories of errors
// bla some preamble if it's useful

#### External process started e.g. `/usr/bin/git --version`
This occurs when you start external processes during the configuration phase. Generally, it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE.

etc.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+3. If not, fix the problems.
+4. Run `./gradlew build --configuration-cache` again. In our experience, this reveals additional issues not seen in the first run.
+
+The problems typically fall into three categories:
+
+1. **external process started `/usr/bin/git --version`**
+2. **cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, a subtype of `org.gradle.api.Project`, as these are not supported with the configuration cache**
+3. **invocation of `Task.project` at execution time is unsupported**
+
+The first error occurs when you start external processes during the configuration phase. Generally, it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE.
```
felixdesouza: This is true, but there's also times where you do need to run things at configuration to decide when/what to configure a la palantir-java-format.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+The problems typically fall into three categories:
+
+1. **external process started `/usr/bin/git --version`**
+2. **cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, a subtype of `org.gradle.api.Project`, as these are not supported with the configuration cache**
+3. **invocation of `Task.project` at execution time is unsupported**
+
+The first error occurs when you start external processes during the configuration phase. Generally, it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE. 
+
+The second and third errors occur because Gradle tasks shouldn't take mutable types as input — `Gradle`, `Settings`, `Project`, `SourceSet`, or `Configuration`. Having these as inputs limits task concurrency (what if two tasks mutate a `Project` concurrently?) and prevents Gradle from serializing task inputs into the cache (they are too large to be serialized).
```
felixdesouza: I don't think size is necessarily an issue here. It's more that you can't reason about the state and reorder and parallelise things if you're reaching into other projects and using that to configure yourself. This is the reason why it limits task concurrency (*within* a project).

It also doesn't want you mutating the big mutable state at task execution time either, you could effectively change how things are and end up with a cache that's inconsistent with reality.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+1. **external process started `/usr/bin/git --version`**
+2. **cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, a subtype of `org.gradle.api.Project`, as these are not supported with the configuration cache**
+3. **invocation of `Task.project` at execution time is unsupported**
+
+The first error occurs when you start external processes during the configuration phase. Generally, it is preferred to run external processes in tasks with properly declared inputs and outputs to avoid unnecessary work when the task is UP-TO-DATE. 
+
+The second and third errors occur because Gradle tasks shouldn't take mutable types as input — `Gradle`, `Settings`, `Project`, `SourceSet`, or `Configuration`. Having these as inputs limits task concurrency (what if two tasks mutate a `Project` concurrently?) and prevents Gradle from serializing task inputs into the cache (they are too large to be serialized). 
+
+To solve the second and third errors, you can
+- Declare the smallest "surface area" your task needs as task input, e.g. take in `Property<String>` for project version, instead of taking in `Project` and accessing `project.version`
```
felixdesouza:
```suggestion
- Declare the smallest "surface area" your task needs as task input, e.g. take in `Property<String>` for the project version, instead of taking in `Project` and accessing `Project#version`
```

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+- [Inject a service](https://docs.gradle.org/current/userguide/service_injection.html) which provides the information/operation you need
+
+
+## Solving Configuration Cache problems
+
+Let’s walk through some examples of fixing Configuration Cache problems
+
+> [!TIP]
+> If you have a huge Gradle projects with many tasks, you can adopt the Configuration Cache incrementally. First, resolve all configuration phase issues, as incremental adoption only works for the execution phase. Then, apply the [gradle-incremental-configuration-cache](https://github.com/palantir/gradle-incremental-configuration-cache) plugin. Gradually add tasks to the allow list as they become compatible.
```
felixdesouza: This seems somewhat out of place:

> First, resolve all configuration phase issues

This is what this whole section is about? Also have we discussed what the difference between configuration phase cc issues and execution phase cc issues are? There's lots to unpack in this tip that is not obvious to me.

This tip probably deserves its own section, at the bottom

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Fixing "external process started"
+
+#### Before: 
+- `MyPlugin` is already a [Gradle managed type](managed-types-and-properties.md)
+- `MyPlugin::apply` starts an external process in the configuration phase, causing the configuration cache to fail
```
felixdesouza:
```suggestion
- `MyPlugin#apply` starts an external process in the configuration phase, causing the configuration cache to fail
```

felixdesouza: explain in a bit more detail what is actually happening here. i.e. `getLatestGitTag` starts an external process calling out to `git`.

felixdesouza: make the example a bit more useful, *use* the result somewhere.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+         throw new RuntimeException(e);
+      }
+   }
+}
+```
+
+#### After
+
+- To safely do an external call during the configuration phase, we can use the `ExecOperations` service.
+- To use `ExecOperations`, we can inject it into `MyPlugin`.
```
felixdesouza: I don't think you want to use `ExecOperations`, that is largely for avoiding issues *within* a Task action i.e. using `project` at execution time. You want to use `ProviderFactory#exec` as that then gives you back a provider that's lazy. You can then pass that around and you may not even execute it! => if it's not on your task graph, you have less things to run to compute your cache key.

If you compare it to some of the work you were doing, we shied away from it for that same reason.

If you have a look at https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:external_processes they make a similar argument i.e. use `providers.exec` or have your own value source that does what you want and gives you a provider back.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+      }
+   }
+}
+```
+
+### Fixing "invocation of `Task.project` at execution time is unsupported"
+
+#### Before
+- `MyTask` is already a [Gradle managed type](managed-types-and-properties.md)
+- `MyTask` is calling `getProject()` at build time, causing the configuration cache to fail: **invocation of `Task.project` at execution time is unsupported**
```
felixdesouza:
```suggestion
- `MyTask` is calling `getProject()` at build time, causing the configuration cache to fail with 

    invocation of `Task.project` at execution time is unsupported
```

let's wrap it in a code block

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+      // Do something with `deleted`...
+      Path projectDir = getProject().getProjectDir().toPath();
+      // Do something with `projectDir`...
+   }
+}
+
+```
+
+#### After
+- To get the project dir, we can use the `ProjectLayout` service.
```
felixdesouza:
```suggestion
- To get the project directory, we can use the `ProjectLayout` service.
```

I don't really like it for code, but understandable, but let's use full names for things when writing docs/prose please.

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+- To do file system operations like delete, we can use the `FileSystemOperations` service
+- We can inject these services into `MyTask`
```
felixdesouza: make a note about the visibility requirements, needing to use `@Inject` and prefixing with `get`. It's a common enough pitfall and the error message sucks when you don't get the correct invocation

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+
+## Two strategies to write Configuration Cache friendly Gradle
+
+To summarize the two strategies we used to solve Configuration Cache issues:
+
+### You can inject services into Gradle-managed types
+
+Do you need the project directory? Inject [`ProjectLayout`](https://docs.gradle.org/current/javadoc/org/gradle/api/file/ProjectLayout.html). Do you need to run a bash command? Inject [`ExecOperations`](https://docs.gradle.org/current/javadoc/org/gradle/process/ExecOperations.html). Gradle offers a [list](https://docs.gradle.org/current/userguide/service_injection.html) of injectable services which will cover 99% of your use cases. However, that list is incomplete — in reality,  almost [anything in Gradle source](https://github.com/search?q=repo%3Agradle%2Fgradle%20%40ServiceScope&type=code) annotated with [`@ServiceScope`](https://github.com/gradle/gradle/blob/196bb409d47f5b6e39d62edd39be939f7606a5cc/platforms/core-runtime/stdlib-java-extensions/src/main/java/org/gradle/internal/service/scopes/ServiceScope.java#L43) can be injected into a Gradle managed type.
```
felixdesouza: * as stated above, the bash command is context dependent, are you in a task action? are you in a plugin code?

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+
+### You can make any class into a Gradle-managed type
+
+A standard class...
```
felixdesouza: You should explain the benefit to doing this, it appears random here.

You haven't actually made any Gradle-managed types in the examples given

guide/adopting-the-configuration-cache.md
```diff
@@ -0,0 +1,384 @@
...
+ 
+    @Inject
+    public Deployment(String customerName) {
+        this.customerName = customerName; 
+    }
+}
+
+```
+
+> [!TIP]
```
felixdesouza: I think you're underselling this!

This probably deserves its own section and/or have most of the examples work via `@Nested`. This is the end goal of where you will probably want to end up in most cases.

guide/adopting-the-configuration-cache.md
felixdesouza: It also means that you don't actually end up talking about the pitfall of the `ProviderFactory#exec` where you need to keep a reference to the returned provider.

felixdesouza (commented):
I have not looked at the errorprone stuff that much, but I would recommend splitting out the docs from the errorprone.

Also spoken elsewhere, have this be single line paragraphs and enable line wrapping in your IDE.

guide/adopting-the-configuration-cache.md
kelvinou01: Yep, let's specify that we only use `ExecOperations` to run bash commands at configuration time.

guide/adopting-the-configuration-cache.md
kelvinou01: Gotcha

guide/adopting-the-configuration-cache.md
kelvinou01: Yep!

guide/adopting-the-configuration-cache.md
kelvinou01: Let's make some gradle-managed types in the examples!

guide/adopting-the-configuration-cache.md
kelvinou01: Yep, let's add a third example, where we explicitly show the benefits of @Nested

kelvinou01:
> I would recommend splitting out the docs from the errorprone

Let's do it 

https://github.com/palantir/gradle-guide/pull/160

CRogers (approved):

guide/adopting-the-configuration-cache.md
kelvinou01: > You probably want to move this section slightly above Configuration Inputs, the main motivation/reasoning is hidden underneath walls of code.

I put this under Configuration Inputs and Configuration Outputs, because they are prerequisites to understanding how CC works. However, I get your point, so I'll add a header to this core motivation section to make it more obvious. 

> just more emphasis on the configuration phase being costly (in and of itself)

Gotcha, I've updated the paragraph. 

> mentioning what tasks requested means e.g. ./gradlew <tasks-requested...> and talking about when you request the same set of tasks etc.

Gotcha, added the distinction to the paragraph.

guide/adopting-the-configuration-cache.md
kelvinou01: This comment clarifies what you mean! Added.

guide/adopting-the-configuration-cache.md
kelvinou01: Good idea!

guide/adopting-the-configuration-cache.md
kelvinou01: Yep!

guide/adopting-the-configuration-cache.md
kelvinou01: True, but maybe that's a little bit too much detail for this guide?

guide/adopting-the-configuration-cache.md
kelvinou01: Gotcha.

guide/adopting-the-configuration-cache.md
kelvinou01: > I don't think you want to use ExecOperations, that is largely for avoiding issues within a Task action i.e. using project at execution time. You want to use ProviderFactory#exec as that then gives you back a provider that's lazy. You can then pass that around and you may not even execute it! 

Thanks for catching this. 

I think i was confused by the wording in Gradle docs (filed [an issue here](https://github.com/gradle/gradle/issues/34293))

But yep, your point is to prefer `ProviderFactory#exec` over other methods like `Project#exec`, `ExecOperations#exec`. Because `ProviderFactory#exec` returns a provider which may or may not be executed, but the other methods execute straight away. 

> It also means that you don't actually end up talking about the pitfall of the ProviderFactory#exec where you need to keep a reference to the returned provider.

Added a [!TIP] box for this!

guide/adopting-the-configuration-cache.md
kelvinou01: > explain in a bit more detail what is actually happening here. i.e. getLatestGitTag starts an external process calling out to git.

Gotcha

> make the example a bit more useful, use the result somewhere.

Hmm, I think adding any more code to the example will make it too complicated. Let's keep it simple, and just use a generic `// Do something...` block?

guide/adopting-the-configuration-cache.md
kelvinou01: > This is what this whole section is about

Hmm, this section talks about both configuration-phase (e.g. external process started) and execution-phase (e.g. calling Task#project in @TaskAction) issues. 

> have we discussed what the difference between configuration phase cc issues and execution phase cc issues are?

I've added a clarification that this refers to problems like `external process started`.

Also, this guide is intended to be standalone, independent of `gradle-incremental-configuration-cache`. I don't think having a separate section for the plugin respects this.

guide/adopting-the-configuration-cache.md
kelvinou01: Got it, removed the inaccurate claim that size was the issue.

PR #173, opened 2025-07-18
Implement the `MethodCallGraph` to catch errorprone violations (e.g. `getProject()`) in helper methods called from `@TaskAction`s

kelvinou01:
## Before this PR

Enhancement to https://github.com/palantir/gradle-guide/pull/142

This PR introduces the `MethodCallGraph`. It will help our error prones catch the following type of Configuration Cache violations:

```java
import org.gradle.api.DefaultTask;
import org.gradle.api.Plugin;
import org.gradle.api.Project;
import org.gradle.api.tasks.TaskAction;

public abstract class PrintProjectName extends DefaultTask {
    private boolean getProjectName(){
        return getProject().getName();
    }
    
    @TaskAction
    public final void action() {
        System.out.println("Project name: " + getProjectName());
    }
}
```

The graph is built on a single compilation unit, that is, it does not cross file boundaries (this is an innate limitation of ErrorProne). Nodes are method symbols. There is an edge from `f` to `g` iff `f` is called by `g` within this compilation unit. It is built on a when-needed basis, i.e. when we encounter a @TaskAction in a file. 

This call graph can be used to enhance other errorprones in the Configuration Cache workflow, and beyond. 

## After this PR

==COMMIT_MSG==
Implement the `MethodCallGraph` to catch errorprone violations (e.g. `getProject()`) in helper methods called from `@TaskAction`s
==COMMIT_MSG==

## Possible downsides?

Performance: For simplicity, `MethodCallGraph ` builds it's graph on a whole compilation unit, rather than incrementally.


gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,488 @@
...
+
+    public static MutableValueGraph<MethodSymbol, Set<MethodInvocationTree>> peekInternalGraph(MethodCallGraph graph)
+            throws NoSuchFieldException, IllegalAccessException {
+        Class<?> methodCallGraphClass = graph.getClass();
+        Field callGraphField = methodCallGraphClass.getDeclaredField("callGraph");
+        callGraphField.setAccessible(true);
+        return (MutableValueGraph<MethodSymbol, Set<MethodInvocationTree>>) callGraphField.get(graph);
+    }
+
+    private VisitorState parseJavaCode(String javaCode) throws IOException {
```
kelvinou01: As discussed in person, let's try to not maintain all this errorprone initialization code. Maybe let's try using the `// BUG:` mechanism to validate our call graph. 

@felixdesouza and I will decide between the current implementation and this proposed one tomorrow, after I finish implementing the proposed one.

kelvinou01:
If a task file contains multiple `@TaskAction`s, the file's call graph will be walked twice. This may influence correctness as well — is an error flagged twice?

gradle-guide-error-prone/build.gradle
```diff
@@ -7,8 +7,10 @@ dependencies {
 
     annotationProcessor 'com.google.auto.service:auto-service'
     compileOnly 'com.google.auto.service:auto-service'
+    implementation 'com.google.errorprone:error_prone_check_api'
 
     testCompileOnly 'org.jetbrains:annotations'
+    testCompileOnly 'com.google.auto.service:auto-service'
```
felixdesouza: I'm surprised you didn't need the `testAnnotationProcessor` for the `auto-service` because then nothing is actually processing the annotations from this class. I think it's probably fine for tests to not have this dependency or annotate your test bug checker with the service annotations considering that you're instantiating it directly perhaps rather than discovering it through service loading.

changelog/@unreleased/pr-173.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: feature
+feature:
+  description:
+    Implement the `CallGraphBugChecker`
```
felixdesouza: Let's give this a bit more meaning here. Not too verbose but something that a user can read and understand the outcome

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+                    consumer.accept(current, neighbor, edgeInvocations, callToChainedCall);
+                }
+            }
+        }
+    }
+
+    private static final Cache<CompilationUnitTree, MethodCallGraph> callGraphCache =
+            Caffeine.newBuilder().maximumSize(1000).weakKeys().build();
+
+    @SuppressWarnings("VisibilityModifier")
```
felixdesouza: this is a bit of a weird thing to suppress, what's the issue here

changelog/1.9.0/pr-160.v2.yml
```diff
@@ -1,6 +0,0 @@
-type: feature
```
felixdesouza: a bit weird that this is in the PR, did you rebase correctly?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+    protected final Supplier<MethodCallGraph> callGraph =
```
felixdesouza: I thought I was going crazy, didn't realise the `Supplier` was an errorprone type :joy:

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+     * A compilation unit is just fancy speak for a java file.
+     */
+    public class MethodCallGraph {
+        /**
+         * Nodes are method declarations.
+         * The edge between two methods f and g are all the instances where f calls g directly.
+         * The edge does not contain any transitive calls from f to g (e.g. f calls h calls g)
+         * For more details, see {@code CallGraphBugCheckerTest} for the graph's specification
+         */
+        @VisibleForTesting // Ideally MethodCallGraph is its own class, but SuppressibleTreePathScanner is protected
```
felixdesouza: idgi, `MethodCallGraph` *is* its own class..?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+
+public abstract class CallGraphBugChecker extends GradleGuideBugChecker {
+    /**
+     * A directed graph of "who calls who, and where" within this compilation unit.
+     * A compilation unit is just fancy speak for a java file.
+     */
+    public class MethodCallGraph {
+        /**
+         * Nodes are method declarations.
+         * The edge between two methods f and g are all the instances where f calls g directly.
```
felixdesouza:
```suggestion
         * The edge between two methods {@code f} and {@code g} are all the instances where {@code f} calls {@code g} directly.
```

similarly for other identifiers in the javadoc pls

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+public abstract class CallGraphBugChecker extends GradleGuideBugChecker {
+    /**
+     * A directed graph of "who calls who, and where" within this compilation unit.
+     * A compilation unit is just fancy speak for a java file.
```
felixdesouza:
```suggestion
     * A directed graph of "who calls who, and where" within this compilation unit (source file).
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.MethodTree;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+public abstract class CallGraphBugChecker extends GradleGuideBugChecker {
```
felixdesouza: this class doesn't seem to do *anything* with the methods from `GradleGuideBugChecker`, doesn't really feel like it should be a bug checker at all

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+                    callGraph.addNode(methodSymbol);
+                    return super.visitMethod(node, Optional.of(methodSymbol));
+                }
+
+                @Override
+                public Void visitMethodInvocation(MethodInvocationTree node, Optional<MethodSymbol> caller) {
+                    MethodSymbol calleeSymbol = (MethodSymbol) ASTHelpers.getSymbol(node.getMethodSelect());
+                    callGraph.addNode(calleeSymbol);
+
+                    if (caller.isPresent()) {
```
felixdesouza: when might you have `caller` be empty? method invocation in a field declaration?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+            }.scan(state.getPath().getCompilationUnit(), Optional.empty());
+
+            new SuppressibleTreePathScanner<Void, Optional<MethodInvocationTree>>(state) {
+                @Override
+                public Void visitMethodInvocation(
+                        MethodInvocationTree call, Optional<MethodInvocationTree> chainedCall) {
+                    chainedCall.ifPresent(chained -> callToChainedCall.put(call, chained));
+                    return super.visitMethodInvocation(call, Optional.of(call));
+                }
+            }.scan(state.getPath().getCompilationUnit(), Optional.empty());
```
felixdesouza: surely you can combine this with the above scan? seems somewhat wasteful to process the same source file *twice*

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+
+                    return super.visitMethodInvocation(node, caller);
+                }
+            }.scan(state.getPath().getCompilationUnit(), Optional.empty());
+
+            new SuppressibleTreePathScanner<Void, Optional<MethodInvocationTree>>(state) {
+                @Override
+                public Void visitMethodInvocation(
+                        MethodInvocationTree call, Optional<MethodInvocationTree> chainedCall) {
+                    chainedCall.ifPresent(chained -> callToChainedCall.put(call, chained));
```
felixdesouza: this map is always 1-1? or can you overwrite things? I'm not too familiar with the types that you get back, but want to make sure

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -73,25 +134,30 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
...
-            .onDescendantOf("org.gradle.api.Project")
-            .named("getLogger");
-
-    public static final String VIOLATION_MESSAGE = "Don't call `getProject()` in task actions";
-
     @Override
     public Description matchMethod(MethodTree tree, VisitorState state) {
-        if (isTaskAction(tree, state) || overridesExecute(tree, state)) {
-            reportAllViolations(state, TASK_GET_PROJECT_METHOD, VIOLATION_MESSAGE);
+        if (shouldTraverseCallGraph(tree, state)) {
```
felixdesouza: let's invert this if, so we have less indentation 🙏

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -73,25 +134,30 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
...
+            MethodSymbol start = ASTHelpers.getSymbol(tree);
+            callGraph.get(state).dfs(start, (from, to, edges, callToChainedCall) -> {
+                for (MethodInvocationTree invocation : edges) {
```
felixdesouza: `edges` is kinda weird, because you don't actually have an edge, you just have a method invocation. You're then forced to construct said edge separately.

Your edge consumer should probably pass in the contents of said edge and the calculation from `callToChainedCall`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -54,8 +54,69 @@
...
+        @Override
+        public void fixOrReport(MethodInvocationTree illegalMethod, MethodInvocationTree chained, VisitorState state) {
+            state.reportMatch(buildDescription(illegalMethod)
+                    .setMessage("Don't call `getProject()` in task actions")
+                    .build());
+        }
+    }
+
+    protected class GetProjectGetLogger implements Violation {
+        /** Matches task.getProject().getLogger(). */
```
felixdesouza: for the docs in the violations, put them in `{@code ...}` blocks please

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -54,8 +54,69 @@
...
+    }
+
+    protected class GetProject implements Violation {
+        /** Matches task.GetProject(). */
+        @Override
+        public boolean matches(MethodInvocationTree illegalMethod, MethodInvocationTree chained, VisitorState state) {
+            return TASK_GET_PROJECT.matches(illegalMethod, state);
+        }
+
+        /** Raises a warning. */
```
felixdesouza: not that informative, I would just remove this

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -54,8 +54,69 @@
...
+
+        /** Fixes task.getProject().getLogger() to task.getLogger(). */
+        @Override
+        public void fixOrReport(MethodInvocationTree illegalMethod, MethodInvocationTree chained, VisitorState state) {
+            SuggestedFix.Builder fix = SuggestedFix.builder();
+            Optional<String> receiverSource =
+                    Optional.ofNullable(ASTHelpers.getReceiver(illegalMethod)).map(state::getSourceForNode);
+            String simplifiedCall =
+                    receiverSource.map(receiver -> receiver + ".").orElse("") + "getLogger()";
+            fix.replace(chained, simplifiedCall).build();
```
felixdesouza: I don't think you need to have a builder saved as a separate argument. It looks like it's enough to basically do:

```
SuggestedFix fix = SuggestedFix.builder().replace(chained, simplifiedCall).build()
```

Then pass that through. You're also calling `build` twice, which there's not much wrong with it, but it is confusing.

My general rule of thumb is to declare variables close to where you're going to use them where possible. Minimises the scope of what the reader may have to think about as they're going through the class.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
felixdesouza: you also don't do anything with `from` and `to` from the looks of it. So it's a little confusing how the `EdgeConsumer` is supposed to be used.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -73,25 +134,30 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
...
     public Description matchMethod(MethodTree tree, VisitorState state) {
-        if (isTaskAction(tree, state) || overridesExecute(tree, state)) {
-            reportAllViolations(state, TASK_GET_PROJECT_METHOD, VIOLATION_MESSAGE);
+        if (shouldTraverseCallGraph(tree, state)) {
+            MethodSymbol start = ASTHelpers.getSymbol(tree);
+            callGraph.get(state).dfs(start, (from, to, edges, callToChainedCall) -> {
+                for (MethodInvocationTree invocation : edges) {
+                    MethodInvocationTree chained = callToChainedCall.get(invocation);
+                    for (Violation violation : violations) {
+                        if (violation.matches(invocation, chained, state)) {
```
felixdesouza: not necessarily actionable right now, but something to think about, i.e. will fixing one thing then introduce another warning? that could be in theory fixed by a later `Violation` class idk. I think there's probalby some suppressible error prone shenanigans that make this okay, but idk

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
felixdesouza: thinking about it a bit more, you don't really care about the `from` and `to` right? That's just a way for you to traverse the graph. The only thing you really care about is a final list of `call -> chained` mappings.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+         * Routine which processes an edge in the call graph.
+         * {@code callToChainedCall} maps a method call to the call immediately chained to it.
+         * We need the chained call to provide auto-fixes, e.g. getProject().getLogger() ==> getLogger()
+         */
+        public interface EdgeConsumer {
+            void accept(
+                    MethodSymbol from,
+                    MethodSymbol to,
+                    Set<MethodInvocationTree> edge,
+                    Map<MethodInvocationTree, MethodInvocationTree> callToChainedCall);
```
felixdesouza: from an earlier comment, this should probably be:

```java
public interface ChainedCallConsumer {
  void accept(MethodInvocationTree call, MethodInvocationTree chained);
}
```

felixdesouza: you then don't really need to expose things like `dfs` and a bunch of other implementation details to the caller. Just a bunch of chained calls instead.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,422 @@
...
+        test(
+                """
+            abstract class TestClass {
+                // BUG: Diagnostic contains: CallGraph: TestClass.action --> [ , ]
+                public abstract void action();
+                }
+        """);
+    }
+
+    @AutoService(BugChecker.class)
```
felixdesouza: yeah I don't think you need this, since it's not being service loaded, it's actually being called directly.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,422 @@
...
+class CallGraphBugCheckerTest {
+    private final CompilationTestHelper compilationTestHelper =
+            CompilationTestHelper.newInstance(TestableCallGraphBugChecker.class, getClass());
+
+    private TestableCallGraphBugChecker checker;
+
+    @BeforeEach
+    void beforeEach() {
+        checker = new TestableCallGraphBugChecker();
+    }
```
felixdesouza: this is unused now since we're not looking at the internal state anymore.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,422 @@
...
+                    String outgoingNode = getFullyQualifiedName(callee).toString();
+                    String diagnostic = String.format("CallGraph Edge: %s", outgoingNode);
+                    state.reportMatch(
+                            buildDescription(tree).setMessage(diagnostic).build());
+                }
+            });
+            return Description.NO_MATCH;
+        }
+
+        public static Name getFullyQualifiedName(MethodSymbol sym) {
```
felixdesouza:
```suggestion
        private static Name getFullyQualifiedName(MethodSymbol sym) {
```

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,422 @@
...
+                }
+
+                // BUG: Diagnostic contains: CallGraph: TestClass.getBuilder --> [ , ]
+                public Builder getBuilder() { return new Builder(); }
+                // BUG: Diagnostic contains: CallGraph: TestClass.getService --> [ , ]
+                public Service getService() { return new Service(); }
+                // BUG: Diagnostic contains: CallGraph: TestClass.getString --> [ , ]
+                public String getString() { return "example"; }
+                // BUG: Diagnostic contains: CallGraph: TestClass.getList --> [ , ]
+                public List<String> getList() { return new ArrayList<>(); }
```
felixdesouza: so I think it's a bit strange to not include constructors. This is then not a generalised call graph

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,422 @@
...
+                    public Object build() { return new Object(); }
+                }
+
+                public static class Chain {
+                    // BUG: Diagnostic contains: CallGraph: TestClass.Chain.second --> [ , ]
+                    public Chain second() { return this; }
+
+                    // BUG: Diagnostic contains: CallGraph: TestClass.Chain.fourth --> [ , ]
+                    public Chain fourth() { return this; }
+                }
```
felixdesouza: I'm a little confused with what you're trying to assert here. Is this just scaffolding for `complexChain` ?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
felixdesouza: hmm, thinking about htis some more, I think generalising like this is probably hurting more than helping. I would probably structure this as:

```java
class ChainedMethodCallCollector {
  // do a bunch of the graph stuff

  // all the chained method calls defined in this method
  Set<ChainedMethodCall> chainedMethodCalls(MethodSymbol method);
}
```

consumer of this:

```java
public Description matchMethod(MethodTree tree, VisitorState state) {
        if (shouldTraverseCallGraph(tree, state)) {
            MethodSymbol start = ASTHelpers.getSymbol(tree);
            Set<ChainedMethodCall> chainedMethodCalls = chainedMethodCallCollector.chainedMethodCalls(start);
            for (ChainedMethodCall chainedMethodCall : chainedMethodCalls) {
              for (Violation violation: violations) {
                // usual stuff
              }
            }
        }

        return Description.NO_MATCH;
    }
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,136 @@
...
+                for (MethodSymbol neighbor : callGraph.successors(current)) {
+                    Set<MethodInvocationTree> edgeInvocations =
+                            callGraph.edgeValue(current, neighbor).orElseGet(Set::of);
+                    consumer.accept(current, neighbor, edgeInvocations, callToChainedCall);
+                }
+            }
+        }
+    }
+
+    private static final Cache<CompilationUnitTree, MethodCallGraph> callGraphCache =
```
felixdesouza: I think we should not bother with this for now, considering we only have one implementation of this `CallGraphBugChecker`. We should just instantiate the call graph cache inside the existing implementation, since currently it will have the same effect.

Should we need to pull it out, we can rethink it again, and I imagine @CRogers may know of a better way to share state across the whole compilation/have ideas of when things will properly go away.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,422 @@
...
+        """);
+    }
+
+    @AutoService(BugChecker.class)
+    @BugPattern(summary = "", severity = SeverityLevel.SUGGESTION)
+    private static final class TestableCallGraphBugChecker extends CallGraphBugChecker
+            implements BugChecker.MethodInvocationTreeMatcher, BugChecker.MethodTreeMatcher {
+        @Override
+        public MoreInfoLink moreInfoLink() {
+            return null;
```
felixdesouza: surprised baseline didn't ding you for this

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecutionTest.java
```diff
@@ -84,6 +84,103 @@ final void action() {
...
+            """,
+                    """
+                import org.gradle.api.DefaultTask;
+                import org.gradle.api.tasks.TaskAction;
+                import org.gradle.api.logging.Logger;
+
+                abstract class CustomTask extends DefaultTask {
+                    @TaskAction
+                    final void action() {
+                        // BUG: Diagnostic contains: Don't call `getProject()` in task actions
```
felixdesouza, line 112: not familiar with the library, but should you not be annotating the before? I suppose this is true for the after as well in this particular case.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecutionTest.java
```diff
@@ -84,6 +84,103 @@ final void action() {
...
+                    """
+                import org.gradle.api.DefaultTask;
+                import org.gradle.api.tasks.TaskAction;
+                import org.gradle.api.logging.Logger;
+
+                abstract class CustomTask extends DefaultTask {
+                    @TaskAction
+                    final void action() {
+                        getProject();
+                        Logger logger = getProject().getLogger();
```
felixdesouza: I think you should annotate what's wrong, so we know what is actually going to be fixed.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -73,25 +134,30 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
...
+                        }
+                    }
+                }
+            });
         }
 
         return Description.NO_MATCH;
     }
 
+    private boolean shouldTraverseCallGraph(MethodTree tree, VisitorState state) {
```
felixdesouza: something to think about for a separate PR, `doLast` and `doFirst` also add task actions. Probably want to check those too

felixdesouza: it's a bit more complicated because you pass in an `Action`, so you have to know to check that at a later point, but that's about it.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: From in person discussion, ChainedMethodCallCollector collects all methods in the transitive closure of a TaskAction

kelvinou01:
As discussed in person, we can put the lazy instantiation of the call graph in a matchCompilationUnit, to keep everything minimally scoped for simplicity.

changelog/1.9.0/pr-160.v2.yml
kelvinou01: As discussed in person, we can use rebase onto, rather than regular rebase (which causes crazy conflicts due to squashing)

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: As discussed in person, let's not suppress `VisibilityModifier`. instead, use a private field and a public getter.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
kelvinou01: Make it more informative "raises a warning because we can't actually fix this"

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: as discussed in person, let's keep this separate to keep it easy to reason with for now, and combine it later if perf is an issue (rather than prematurely optimize)

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
kelvinou01: As discussed in person, we already catch Action<Task>, so we're already checking doFirst and doLast. 

Let's add a test to check that doFirst and doLast is caught, both as a standalone class as well as an anonymous class.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: Hmmph, the only reason I created this subclass is that `MethodCallGraph` needs `SuppressibleTreePathScanner`. `SuppressibleTreePathScanner` is unfortunately a protected field of `BugChecker` (which makes sense as suppressions are per-bug checker), hence `MethodCallGraph` has to be defined in a subclass of the `BugChecker`

I'd much rather have `MethodCallGraph` be a standalone class, making it easier to test and read. However, I don't see a good way to inject `SuppressibleTreePathScanner` — we can't do dynamic subclassing

```java

class MethodCallGraph {
    public MethodCallGraph(VisitorState state, Class<TreeVisitor> treeVisitorClass) {
        // doesn't compile!!
        treeVisitorClass.getDeclaredConstructor(VisitorState.class).newInstance(state) {
            @Override
            public Void visitMethod(MethodTree node, Optional<MethodSymbol> caller) {
                MethodSymbol methodSymbol = ASTHelpers.getSymbol(node);
                callGraph.addNode(methodSymbol);
                return super.visitMethod(node, Optional.of(methodSymbol));
            }

            @Override
            public Void visitMethodInvocation(MethodInvocationTree node, Optional<MethodSymbol> caller) {
                MethodSymbol calleeSymbol = (MethodSymbol) ASTHelpers.getSymbol(node.getMethodSelect());
                callGraph.addNode(calleeSymbol);

                if (caller.isPresent()) {
                    MethodSymbol callerSymbol = caller.get();
                    Set<MethodInvocationTree> existingInvocations =
                            callGraph.edgeValue(callerSymbol, calleeSymbol).orElseGet(HashSet::new);
                    existingInvocations.add(node);
                    callGraph.putEdgeValue(callerSymbol, calleeSymbol, existingInvocations);
                }

                return super.visitMethodInvocation(node, caller);
            }
        }.scan(state.getPath().getCompilationUnit(), Optional.empty())
   ...
   }
...
}
```

We could also pass `SuppressibleTreePathScanner` as a constructor arg. But all the code that defines how we construct the graph will lie outside of the MethodCallGraph. This isn't ideal. 

```java
class MethodCallGraph {
    public MethodCallGraph(VisitorState state, TreeVisitor treeVisitor) {
        visitor.scan(state.getPath().getCompilationUnit(), Optional.empty());
    }
}

// All the tree scanning code is in the caller of `new MethodCallGraph(..)`
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: sorry by it's own class i mean "not a nested class". see https://github.com/palantir/gradle-guide/pull/173/files#r2235647488

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: I thought about this more over the weekend.

1. Fundamentally, `CallGraphBugChecker` builds a data structure, then processes that data structure. Having the public API reflect this is intuitive.

   "A list of calls under the transitive closure of a @TaskAction" is nice to work with upfront —just iterate and process. But when you need to understand the system "what is this actually doing?", you have to jump thru a hoop to eventually arrive at the mental model of "graph, then DFS" anyways. The OOP wrapper hides this reality and creates unnecessary cognitive overhead.

   The graph approach is also more flexible. We can easily add different traversal algorithms or edge consumers without changing the core structure. A `ChainedCallCollector` would lock us into one specific pattern.

2. On the `EdgeConsumer` interface—I think a consumer that takes two nodes and the edge between them is fairly intuitive. It maps directly to how you think about graph operations. Having a `ChainedCallCollector` feels too much like pandering the parent class to its consumers. 

What do you think?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: Hmm, we do need some caching using the compilation unit as the key, and the built call graph as a value. This is because if a compilation unit contains two `@TaskAction`s, `IllegalMethodCalledDuringTaskExecution` will be matched twice, hence building the call graph twice.

However, I realize there is a separate subtlety here. Let's say in the future we have a second subclass of `CallGraphBugChecker` — `ProcessRunDuringConfigurationTime`. Then, `ProcessRunDuringConfigurationTime` and `IllegalMethodCalledDuringTaskExecution`  cannot share the same call graph cache. This is because building the call graph on `ProcessRunDuringConfigurationTime` and `IllegalMethodCalledDuringTaskExecution` may yield different call graphs, as [we may suppress one without the other](https://github.com/google/error-prone/blob/249c359d98349107b045f5de6f06c3098caf2c76/check_api/src/main/java/com/google/errorprone/bugpatterns/BugChecker.java#L324).

I'll leave out the caching for now then, let's address this in the future.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
kelvinou01: Thanks, this is a good idea. I'll move the variables close to where they're used.

changelog/@unreleased/pr-173.v2.yml
kelvinou01: Yep, let's do "Implement the `CallGraphBugChecker` to catch errorprone violations in helper methods"

gradle-guide-error-prone/build.gradle
kelvinou01: Thanks for catching this. 

I added @AutoService in the test because gradle-baseline's [BugCheckerAutoService](https://github.com/palantir/gradle-baseline#baseline-error-prone-checks) would complain otherwise. 

Seems to me that suppressing BugCheckerAutoService in the test is the correct thing to do, rather than adding a test dependency on auto-service. I'll also add a comment explaining the suppression.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: Caller is empty when we initially start scanning — a CompilationUnit has no caller. 

I thought using Optional to model this fact was nicer than nullchecking.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: Yep, this map is always 1-1.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
kelvinou01: Yes!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
kelvinou01: See https://github.com/palantir/gradle-guide/pull/173#r2235880563 for a full response to this

> On the EdgeConsumer interface—I think a consumer that takes two nodes and the edge between them is fairly intuitive. It maps directly to how you think about graph operations. Having a ChainedCallCollector feels too much like pandering the parent class to its consumers. Some degree of that is inevitable (like callToChainedCall), but having the fundamental structure of the public API reflect its consumer is a bit too much.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecutionTest.java
kelvinou01: Nope, the before doesn't have to be annotated with `BUG: Diagnostic contains:` here. Only the after has to be annotated.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecutionTest.java
kelvinou01: Yep! That'd make reading the tests abit easier.

changelog/1.9.0/pr-160.v2.yml
kelvinou01: https://andrewlock.net/working-with-stacked-branches-in-git-is-easier-with-update-refs/

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: yep!

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
kelvinou01: Yep!

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
kelvinou01: `TestableCallGraphBugChecker` requires that you declare the complete set of edges for each method declaration. We need this to test the call graph completely.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
kelvinou01: Hmmph, good catch. Seems like constructor invocations [are a special kind of expression](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.compiler/com/sun/source/tree/NewClassTree.html), distinct to method invocations. 

I tried drafting a version that includes constructor invocations. It introduces quite a bit of (albeit intrinsic) complexity to both the graph and the violation handling code, as we have to switch on either method invocation VS constructor invocation when processing an edge. We could include it in a future PR if needed.

```java
    protected class CallGraph {

        public sealed interface Call permits MethodCall, ConstructorCall { }

        public record MethodCall(MethodInvocationTree invocation)
                implements Call {}

        public record ConstructorCall(NewClassTree construction)
                implements Call {}

        @VisibleForTesting
        protected final MutableValueGraph<MethodSymbol, Set<Call>> callGraph =
                ValueGraphBuilder.directed().allowsSelfLoops(true).build();
        ...
    }
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
felixdesouza: I think we discussed the former point no? i.e. do the collection when you visit the compilation unit. Save it to a field (although I'm not sure what the lifecycle is here) or attach it to the state that you pass around, so then it's cached for remainder of the compilation unit.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,139 @@
...
+     */
+    protected class MethodCallGraph {
+
+        /**
+         * Represents a method call with optional chained call information.
+         * For example, in {@code getProject().getLogger()}, {@code rootCall = getProject()},
+         * {@code chainedCall = getProject().getLogger()}
+         *
+         */
+        public record MethodCall(MethodInvocationTree rootCall, Optional<MethodInvocationTree> chainedCall) {}
```
kelvinou01, line 46: FYSA ive embedded information about the chained calls into the call graph itself. This was done to [move the complexity from the code to the data structure](https://softwareengineering.stackexchange.com/q/163185). 

**Before**: Although the graph was a simpler data structure,

```java
protected final MutableValueGraph<MethodSymbol, Set<MethodInvocationTree>> callGraph =
        ValueGraphBuilder.directed().allowsSelfLoops(true).build();
```

...we needed to keep this map to get chained calls (e.g. getProject() --> getProject().getLogger())...

```java
private final Map<MethodInvocationTree, MethodInvocationTree> callToChainedCall = new HashMap<>();
```

...resulting in the `EdgeConsumer` being a leaky abstraction.

```java
public interface EdgeConsumer {
    void accept(
        MethodSymbol from,
        MethodSymbol to,
        Set<MethodInvocationTree> edge,
        Map<MethodInvocationTree, MethodInvocationTree> callToChainedCall);
}
```

**After**: Each MethodCall now contains it's own chained call information:

```java
public record MethodCallEdge(
    MethodInvocationTree invocation, 
    Optional<MethodInvocationTree> chainedCall) {
}
```

...we embed the complexity in the data structure itself...

```java
protected final MutableValueGraph<MethodSymbol, Set<MethodCall>> callGraph =
      ValueGraphBuilder.directed().allowsSelfLoops(true).build();
```

... the result is a clean, self-contained interface with all necessary information embedded in the edge.

```java
public interface EdgeConsumer {
      void accept(MethodSymbol from, MethodSymbol to, Set<MethodCall> edge);
}
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: https://github.com/palantir/gradle-guide/pull/173#issuecomment-3119037918

kelvinou01: Sorry let me revisit this. This PR is getting quite large so I must've missed this note I made from large week 😵‍💫

kelvinou01: Alright i've tested the current way we're caching call graphs, and verified that 

- It only builds the graph once, even with multiple `@TaskAction`s.
- if there's no `@TaskAction`s, it doesn't build. 

```java
    private MethodCallGraph callGraph;

    /**
     * The first call to this method builds the call graph on the state's compilation unit.
     * Subsequent calls will return the cached call graph.
     */
    protected MethodCallGraph getCallGraph(VisitorState state) {
        if (callGraph == null) {
            callGraph = new MethodCallGraph(state);
        }

        return callGraph;
    }
```

So using matchCompilationUnit shouldn't be needed? (Also, there is no defined ordering between matchCompilationUnit and matchMethod)

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
kelvinou01: We have decided to implement the test [using comments to assert the graph's edges](https://github.com/palantir/gradle-guide/blob/6f58c5e6f146c6f7de1ef1a7954ef41d9079d73a/gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java), rather than [snooping on the internal `MutableValueGraph` and asserting on it's edges](https://github.com/palantir/gradle-guide/blob/feff3607b9fa0170b3b5b7ba61d5e0614207c666/gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java)

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
kelvinou01: Yep!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
felixdesouza: Core tenet for me is: I think it's still too early to tell what the generalised form is supposed to look like without any real examples. 

I wouldn't go as far as to say it was intuitive, at least for me. Between the caching, the separate class for the call graph, the `EdgeConsumer` where the nodes weren't used but the edges were. It was non-trivial to grok/see what you were generally trying to do. I also don't think things need to be super flexible either. Things can change at later points, and right now I'm not necessarily convinced that this is the shape you want for future stuff based on the existing sample size of 1.

Re `EdgeConsumer`, it is also not "general" either! It's passed a map for some chained calls which is a pretty weird abstraction. I think my main gripe is largely flip flopping/indirection between the classes because one class knows how to do certain bits i.e. the dfs belongs in the call graph, and the edge consumer is doing other bits (processing something from the call graph, when we can actually just make the call useful in one go. It also does not help that in example given, you don't actually use two of the parameters.

I've had some time to think about it, and poking around the API a bit, and I think you can probably make this simpler:

```java
interface CallGraph {
  Set<MethodSymbol> allCallers(MethodSymbol method);
}
```

This is somewhat the interface you get from IntelliJ's "Call Hierarchy" tab - there's also callees, but I prefer the callers approach. You could probably go one step further and include more information e.g. locations where it's transitively called, but I don't think it's that important rn.

How would you consume this? Instead of matching on the `MethodTree`, why not instead:
* match on the `MethodInvocationTree`
  * `ASTHelpers#getReceiver(methodTree)`, gets you the tree without having to do a separate full compilation unit scan. I believe it's here you use your `Violation#matches`
* find the enclosing method
  * `ASTHelpers#findEnclosingMethod`
* then find all the callers of the method. 
  * `callGraph.allCallers(symbol)
* You can then check whether any of those methods are `TaskAction`s and then you're done.
  * `ASTHelpers#findMethod(symbol, state)` - gives you back a method tree
  * use your matcher for `TaskAction` or `execute` or whatever.

Why do I think this works a bit better?
* You're matching on elements you care about i.e. is this bad method invocation called by a task action
* you can quickly check whether the invocation is something you want to check
* you consult the call graph for the information you want, you deal with the information right away instead of indirection
* graph construction is also easier and less convoluted

I still think it's probably too early to make such an interface without at least a couple more use cases and would still prefer just having a very non-general method that does what we want. But if you feel so strongly about it, this is a reasonable-ish middle ground.

kelvinou01: > I wouldn't go as far as to say it was intuitive, at least for me......I'm not necessarily convinced that this is the shape you want for future stuff based on the existing sample size of 1.

Got it, thanks for the feedback. I was too in the weeds to notice that things were getting convoluted. Let's make the interface of `CallGraphBugChecker` more specialized to our current use case

> Re EdgeConsumer, it is also not "general" either! It's passed a map for some chained calls which is a pretty weird abstraction. I think my main gripe is largely flip flopping/indirection between the classes because one class knows how to do certain bits 

Got it. I've now embedded information about the chained call directly into the graph's edges. `callsUnderTransitiveClosureOf(MethodSymbol)` returns a set of `record MethodCall(MethodInvocationTree rootCall, Optional<MethodInvocationTree> chainedCall)`. Hopefully this reduces the amount of flipping it needs. 

Also I agree that the old `EdgeConsumer` which took in the mapping `callToChainedCall` was leaky and bad.

kelvinou01: > ...locations where it's transitively called, but I don't think it's that important rn.

> why not instead: 
> - match on the MethodInvocationTree
> - find the enclosing method

Okay I tried matching on `MethodInvocationTree`, and it makes the call graph so much simpler! Now we don't even need a ValueGraph. 

Feels like we should've done this from the beginning 🙃 Thank you for suggesting this!

gradle-guide-error-prone/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
 
     annotationProcessor 'com.google.auto.service:auto-service'
     compileOnly 'com.google.auto.service:auto-service'
+    implementation 'com.google.errorprone:error_prone_check_api'
```
CRogers: This is already in `api` - do we need it as part of `implementation`?

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
```diff
@@ -0,0 +1,214 @@
...
+import org.intellij.lang.annotations.Language;
+import org.junit.jupiter.api.Test;
+
+@SuppressWarnings("LineLength")
+class CallGraphBugCheckerTest {
+    private final CompilationTestHelper compilationTestHelper =
+            CompilationTestHelper.newInstance(TestableCallGraphBugChecker.class, getClass());
+
+    @Test
+    void testDirectCallsOnly_NoTransitiveCalls() {
```
CRogers: Our style is not to prefix test methods with `test` and to write the tests as english sentences in snake_case.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,100 @@
...
+         * {@code g}.
+         */
+        @VisibleForTesting
+        protected final ImmutableGraph<MethodSymbol> callGraph;
+
+        public MethodCallGraph(VisitorState state) {
+            MutableGraph<MethodSymbol> mutableCallGraph =
+                    GraphBuilder.directed().allowsSelfLoops(true).build();
+
+            new SuppressibleTreePathScanner<Void, Optional<MethodSymbol>>(state) {
```
CRogers: Does this mean that a `@SuppressWarnings` will stop us from finding whole subgraphs because "on the way" something was suppressed? ie, here we would not even visit `anotherMethod` because `suppressed` has been suppressed?

```java
@TaskAction
void action() {
    suppressed();
}

@SuppressWarnings("IllegalMethodCalledDuringTaskExecution")
void suppressed() {
    anotherMethod();
}

void anotherMethod() {
    getProject();
}
```

I feels weird that we would not consider `anotherMethod` as it's not suppressed anywhere along it's `TreePath`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
```diff
@@ -0,0 +1,100 @@
...
+            return result;
+        }
+    }
+
+    private MethodCallGraph callGraph;
+
+    /**
+     * The first call to this method builds the call graph on the state's compilation unit.
+     * Subsequent calls will return the cached call graph.
+     */
```
CRogers: Error Prone [keeps an instance of each `BugChecker` around for the entirety of the compilation stage](https://github.com/google/error-prone/blob/c59a1c9bdac1fc4f35e5f842ab537cb6c5d594e0/check_api/src/main/java/com/google/errorprone/scanner/ErrorProneScanner.java#L176), it doesn't make one per compilation unit. I think this means that this call graph will be created on the first compilation unit it sees then reused on later ones?

I think we need to use a `WeakMap<CompilationUnitTree, MethodCallGraph>` here so we can just store them all in a map, then when the compilation unit is finished/garbage collected, it automatically gets removed from the map to not blow up memory usage.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
CRogers: I think if we don't need to use `SuppressibleTreePathScanner` we don't need to make this a subclass of `BugChecker` and it can just be a static utility.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -54,8 +55,100 @@
...
+    interface Violation {
+        /** If we encounter {@code illegalCall}. */
+        boolean matches(MethodInvocationTree illegalCall, VisitorState state);
+
+        /** Then, suggest fixes or a warning. */
+        void fixOrReport(MethodInvocationTree illegalCall, VisitorState state);
+    }
+
+    protected class GetProject implements Violation {
+        /** Matches {@code task.GetProject()}. */
```
CRogers: `GetProject` -> `getProject`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
CRogers: We need some test for the `SuppressWarnings` cases at least

gradle-guide-error-prone/build.gradle
kelvinou01: You're right, it works without this dependency. Let's remove it.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/CallGraphBugCheckerTest.java
kelvinou01: yep!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: Hmm good suggestion! Conveniently BugChecker already uses `SuppressibleTreePathScanner`, we don't have to do SuppressibleTreePathScanner ourselves

This has allowed me to make the `MethodCallGraph` it's own class independent of the `BugChecker`. 

I've also added [a test](https://github.com/palantir/gradle-guide/pull/173/files#diff-29ad2989688f02530e9029c303ad7215d0af94dd532926de4a1627ab0113ecc1R225) for the "suppressed on the way" case you described above.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/CallGraphBugChecker.java
kelvinou01: Good idea, but let's use a caffeine cache with weak keys and soft values? Their `LoadingCache` has nice, single-source-of-truth semantics.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -55,7 +60,108 @@
         the tasks become independent and can execute in parallel.
         """)
 public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBugChecker
-        implements BugChecker.MethodTreeMatcher {
+        implements BugChecker.MethodInvocationTreeMatcher {
+    private static final LoadingCache<CompilationUnitTree, MethodCallGraph> callGraphCache =
+            Caffeine.newBuilder().maximumSize(1_000).weakKeys().softValues().build(MethodCallGraph::new);
```
CRogers: We should put this in the method call graph - if multiple bugcheckers want access to the method call graph for the same compilation unit, they should use the same cache.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/utils/MethodCallGraph.java
```diff
@@ -0,0 +1,82 @@
...
+ * A directed graph of "who is called by who" within this compilation unit (source file).
+ */
+public class MethodCallGraph {
+    /**
+     * Nodes are {@code MethodSymbol}s. There is an edge from {@code f} to {@code g} iff {@code f} is called by
+     * {@code g}.
+     */
+    private final ImmutableGraph<MethodSymbol> callGraph;
+
+    public MethodCallGraph(CompilationUnitTree compilationUnitTree) {
```
CRogers: We should make the constructor private so that we _force_ bugcheckers to get the cached instance - no chance they can make new ones of these that aren't shared.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecutionTest.java
```diff
@@ -107,6 +204,49 @@ public final void action() {
...
+
+                    @SuppressWarnings("IllegalMethodCalledDuringTaskExecution")
+                    void no_oversuppression() {
+                        helper();
+                    }
+
+                    void helper() {
+                        // BUG: Diagnostic contains: Don't call `getProject()` in task actions
+                        String projectPath = getProject().getPath();
+                    }
```
CRogers, line 304: nice

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -55,7 +60,108 @@
...
+            .named("getLogger");
+    private static final Matcher<ExpressionTree> TASK_GET_PROJECT = MethodMatchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.Task")
+            .named("getProject");
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (isSuppressed(tree, state)) {
+            return Description.NO_MATCH;
+        }
```
CRogers: error-prone should do this for us - no need to do it again.

CRogers: As in, it won't even call this BugChecker if it's suppressed on this tree.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -55,7 +60,108 @@
...
+                .filter(violation -> violation.matches(tree, state))
+                .findFirst();
+        if (firstMatchingViolation.isEmpty()) {
+            return Description.NO_MATCH;
+        }
+
+        CompilationUnitTree compilationUnit = state.getPath().getCompilationUnit();
+        MethodCallGraph callGraph = callGraphCache.get(compilationUnit);
+        MethodSymbol matched = ASTHelpers.getSymbol(tree);
+        Set<MethodSymbol> transitiveCallers = callGraph.transitiveCallers(matched);
```
CRogers, line 82: I wonder whether we should weakly cache this too - idk how expensive the search is, but if it's being called multiple times per method it might make sense.

Then again caching too much is also bad. I say we leave it for now but bear it mind if there is a performance issue with this check.

CRogers: The only repos who get the performance cost is those with bad code I guess, and all these will be removed, so even less reason to care.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/IllegalMethodCalledDuringTaskExecutionTest.java
CRogers, line 1: Can we add a test cases for:
* Explicitly that we can call `getProject()` the constructor
* A method in a `Task` class that is not called by anything else can call `getProject()`
* A method in a `Task` class that is called only by the constructor can call `getProject()`
* A method called by both the constructor and within a `@TaskAction` cannot call `getProject()`

kelvinou01: > Explicitly that we can call getProject() the constructor

Added `getProject_in_constructors_should_pass`, which _specifies_ that stuff in the transitive closure of the constructor shouldn't be flagged by this errorprone

> A method in a Task class that is not called by anything else can call getProject()

Yep, it's already in `getProject_outside_of_TaskAction_should_pass`

> A method in a Task class that is called only by the constructor can call getProject()

Included that case in `getProject_in_constructors_should_pass` 

> A method called by both the constructor and within a @TaskAction cannot call getProject()

Added `getProject_in_shared_helper_should_fail`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/utils/MethodCallGraph.java
kelvinou01: Yep. I wasn't sure if the `MethodCallGraph` should concern itself with caching but this is a good way to prevent perf regressions.

CRogers (approved):

PR #198, opened 2025-08-21
Autofix non-abstract tasks

kelvinou01:
## Before this PR

We want to autofix lots of configuration cache violations in tasks by injecting services. To inject services, tasks have to be declared as abstract.

## After this PR

This PR introduces an autofix from non abstract `Task` to abstract `Task`.

Undoubtably, this will save us [**a lot** of work](https://pl.ntr/non-abstract-tasks). However, this autofix will introduce an API break for all public non-abstract tasks. For example, 

#### Producing repository: Autofix merges in successfully to make `MyTask` abstract

```java
public abstract class MyTask extends DefaultTask { ... }
```

#### Consuming repository: blocked from receiving updates of producing repository due to CI failure caused by manual instantiation

```java
...
MyTask task = new MyTask();
...
```

This should be fine, as usually tasks are not instantiated manually. 

However, we might want to have the change be an autorelease, or have some sort of manual approval by repo owners. 

==COMMIT_MSG==
Autofix non-abstract tasks
==COMMIT_MSG==

## Possible downsides?

FLUPS: should we fix the following as well:
- Projects: should be easy
- Extensions: abit trickier, as they are POJOs


CRogers:
Another key part to note here is that [Gradle 9 requires types that extends Gradle types to be abstract](https://docs.gradle.org/9.0.0/userguide/upgrading_major_version_9.html#injection_getters_are_now_abstract) if they use `@Inject`

CRogers (approved):

PR #210, opened 2025-09-01
use `GradleExec#exec` over `ProviderFactory#exec` 

FinlayRJW:
## Before this PR
Had no errorprone to encourage the usage of `GradleExec`

## After this PR
We now have an errorprone to force the usage of `GradleExec#exec` with updated `README`

See https://github.com/palantir/gradle-utils/pull/435#discussion_r2309876653 for reference 

==COMMIT_MSG==
use `GradleExec#exec` over `ProviderFactory#exec` 
==COMMIT_MSG==


guide/adopting-the-configuration-cache.md
```diff
@@ -183,8 +183,8 @@ abstract class MyPlugin implements Plugin<Project> {
 
 #### After
 
-- To safely do an external call during the configuration phase, we can use the `ProviderFactory` service.
-- To use `ProviderFactory`, we can inject it into `MyPlugin`. 
+- To safely do an external call during the configuration phase, we can use the  [`GradleExec`](https://github.com/palantir/gradle-utils?tab=readme-ov-file#gradleexec) service.
+- To use `GradleExec`, we can inject it into `MyPlugin`.
```
kelvinou01, line 187:
```suggestion
- To use `GradleExec`, we can inject it into `MyPlugin`. 

> [!NOTE]
> [`GradleExec`](https://github.com/palantir/gradle-utils?tab=readme-ov-file#gradleexec) is our version of `ProviderExec` with better error handling and readable stack traces. 
```

guide/adopting-the-configuration-cache.md
```diff
@@ -195,14 +195,13 @@ abstract class MyPlugin implements Plugin<Project> {
 
 ```java
 abstract class MyPlugin implements Plugin<Project> {
-   @Inject
-   protected abstract ProviderFactory getProviderFactory();
+   @Nested
+   protected abstract GradleExec getGradleExec();
```
kelvinou01, line 202: Let's test that this compiles!

kelvinou01 (commented):
🎉
I think this is good to go. Let's wait for a review from @CRogers before we merge.

guide/adopting-the-configuration-cache.md
FinlayRJW: should do thats how its used currently - https://github.com/palantir/gradle-utils/blob/develop/platform/src/main/java/com/palantir/platform/GradleOperatingSystem.java#L33

FinlayRJW:
marked as DNM until we settle how we want to handle errors (i.e https://github.com/palantir/gradle-utils/pull/435) so we don't roll this out only to have to update later

PR #213, opened 2025-09-04
Autofix `getProject().getProjectDir()` to `getProjectLayout().getProjectDirectory().getAsFile()`

kelvinou01:
## Before this PR

PRs in order: https://github.com/palantir/gradle-guide/pull/214 ➡️ https://github.com/palantir/gradle-guide/pull/215 ➡️ https://github.com/palantir/gradle-guide/pull/213 (curr)

We are rolling out the configuration cache. We are doing this [incrementally, one task at a time](https://github.com/palantir/gradle-incremental-configuration-cache). However, for incremental rollout to begin, [all configuration cache problems that occur in the configuration phase must be fixed](https://github.com/palantir/gradle-incremental-configuration-cache/pull/6). 

We've added an [errorprone](https://github.com/palantir/gradle-guide/pull/142) to prevent people from using `getProject` during the execution time. In most cases, when people do `getProject().X()` during execution time,  we can substitute it for another method which is configuration cache friendly — see these [metrics](https://pl.ntr/2vj). `getProject().getProjectDir()` is one such case, let's autofix these!

## After this PR

==COMMIT_MSG==
Autofix `getProject().getProjectDir()` to `getProjectLayout().getProjectDirectory().getAsFile()`
==COMMIT_MSG==

PR #214, opened 2025-09-04
Refactor `IllegalMethodCalledDuringTaskExecution` and it's tests

kelvinou01:
## Before this PR

PRs in order: https://github.com/palantir/gradle-guide/pull/214 (curr) ➡️ https://github.com/palantir/gradle-guide/pull/215 ➡️ https://github.com/palantir/gradle-guide/pull/213 

https://github.com/palantir/gradle-guide/pull/213 is huge — let's split up the refactoring from the actual code changes (213). 

Also, there will be ~a dozen `Violation`s planned to be added, so let's create a separate `taskexecution` package for those to live in. 

## After this PR

- Split the tests into their own classes
- Introduce util class for shared variables and helpers
- Made some snake_case variables camelCase
- Fixed ungrammatical error message

No user-facing behavior changes are introduced. 

==COMMIT_MSG==
Refactor `IllegalMethodCalledDuringTaskExecution` and it's tests
==COMMIT_MSG==


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/utils/Tasks.java
```diff
@@ -0,0 +1,98 @@
...
+    public static final Matcher<ExpressionTree> PROJECT_GET_LOGGER = MethodMatchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.Project")
+            .named("getLogger");
+    public static final Matcher<ExpressionTree> TASK_GET_PROJECT = MethodMatchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.Task")
+            .named("getProject");
+
+    // Optional.empty() in projects without gradle on the classpath
+    // In that case, we should `return Description.NO_MATCH`
+    public static final Supplier<Optional<ClassSymbol>> ACTION_SYM = VisitorState.memoize(
```
CRogers: Let's avoid shortening names - I'd call this `ACTION_SYMBOL` instead (and same for the rest)

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -0,0 +1,104 @@
...
+                });
+        if (isInvokedAtTaskExecution) {
+            firstMatchingViolation.get().fixOrReport(tree, state);
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    // Only the first matching violation is applied. Put more specific violations first.
+    private final List<Violation> violations = List.of(new GetProjectGetLogger(this), new GetProject(this));
```
CRogers: Let's put this at the top of the class - it's unusual to have fields anywhere other than the top of the class.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/violations/GetProject.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.google.errorprone.VisitorState;
+import com.palantir.gradle.guide.errorprone.taskexecution.IllegalMethodCalledDuringTaskExecution;
+import com.palantir.gradle.guide.errorprone.utils.Tasks;
+import com.sun.source.tree.MethodInvocationTree;
+
+public final class GetProject implements Violation {
+    private final IllegalMethodCalledDuringTaskExecution illegalMethodCalledDuringTaskExecution;
+
+    public GetProject(IllegalMethodCalledDuringTaskExecution illegalMethodCalledDuringTaskExecution) {
```
CRogers: Rather than injecting `IllegalMethodCalledDuringTaskExecution` into every `Violation` class, can we just give a `Description` in `fixOrReport` that has already been created using `state.reportMatch(illegalMethodCalledDuringTaskExecution.buildDescription(illegalCall)`? I think we could avoid needing the `tree` or the `state` in `fixOrReport` then and just have it return a `Description`.

CRogers: This assumes that if `matches` returns true, the violation
 will _definitely_ produce a fix or report.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/violations/Violation.java
```diff
@@ -0,0 +1,32 @@
...
+package com.palantir.gradle.guide.errorprone.taskexecution.violations;
+
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.sun.source.tree.MethodInvocationTree;
+
+/**
+ * Represents strategies to report or auto-fix illegal methods and methods chained to it.
+ */
+public sealed interface Violation permits GetProject, GetProjectGetLogger {
```
CRogers: `Violation` is a very generic name here - perhaps we should call it `IllegalMethodViolation` to make it clear it's related to the check.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/utils/Tasks.java
kelvinou01: Yep!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/violations/Violation.java
kelvinou01: Yep! How about `TaskExecutionViolation`?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/violations/GetProject.java
kelvinou01: > I think we could avoid needing the tree or the state in fixOrReport then and just have it return a Description

In general I don't think it's possible to achieve a purely functional interface here. There's no running away from passing the `VisitorState` class, as it is needed for various AST Helper functions. e.g. to render a fix, we need `state.getSourceForNode`.

kelvinou01: Also, from in-person discussion, there is a world where we define a a DSL for Gradle errorprones, so that specific violations can just do

```java
public final class GetProjectGetProjectDir implements TaskExecutionViolation {
    ....
    public GradleFix tryFix(methodTree, state) {
        ...
        return GradleFix.of(methodTree, state)
            .turnAbstract()
            .inject(ProjectLayout.class)
            .replace(methodTree, fixedCall);
    }
}
``` 

But right now we only have 3 errorprones in this domain, including the one in flight at #213. It's pretty hard to foresee the correct API now. Let's go with a simple approach for now, and evolve the API/make a DSL when it makes sense, after a few more errorprones have been added?

kelvinou01: > This assumes that if matches returns true, the violation
will definitely produce a fix or report.

yep, this is the intended behavior.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecutionTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+public abstract class IllegalMethodCalledDuringTaskExecutionTest {
+    private final CompilationTestHelper compilationTestHelper =
+            CompilationTestHelper.newInstance(IllegalMethodCalledDuringTaskExecution.class, getClass());
+
+    protected final void test(@Language("Java") String source) {
+        compilationTestHelper.addSourceLines("Test.java", source).doTest();
+    }
+
+    protected final void test_fix(String filename, @Language("Java") String before, @Language("Java") String after) {
```
CRogers: Call this `testFix` - keep snake_case only for test names

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/violations/GetProjectGetLogger.java
```diff
@@ -0,0 +1,60 @@
...
+
+        Optional<ExpressionTree> receiverMaybe = Optional.ofNullable(ASTHelpers.getReceiver(illegalCall));
+        boolean receiverIsMethodCall = receiverMaybe
+                .map(receiver -> receiver instanceof MethodInvocationTree)
+                .orElse(false);
+        if (receiverIsMethodCall) {
+            MethodInvocationTree receiverCall = (MethodInvocationTree) receiverMaybe.get();
+            return Tasks.TASK_GET_PROJECT.matches(receiverCall, state)
+                    && Tasks.PROJECT_GET_LOGGER.matches(illegalCall, state);
+        }
```
CRogers: A nicer way to write this would be something like

```java
ExpressionTree receiver = ASTHelpers.getReceiver(illegalCall));
if (receiver == null) {
    return false;
}

if (receiver instanceof MethodInvocationTree methodInvocationTree) {
    // just use methodInvocationTree directly.
}

return false;
```

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecutionTest.java
kelvinou01: Yep! I've fixed up the other files as well.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -60,13 +55,8 @@
...
         implements BugChecker.MethodInvocationTreeMatcher {
-
-    private static final Matcher<ExpressionTree> PROJECT_GET_LOGGER = MethodMatchers.instanceMethod()
-            .onDescendantOf("org.gradle.api.Project")
-            .named("getLogger");
-    private static final Matcher<ExpressionTree> TASK_GET_PROJECT = MethodMatchers.instanceMethod()
-            .onDescendantOf("org.gradle.api.Task")
-            .named("getProject");
+    // Only the first matching violation is applied. Put more specific violations first.
+    private final List<Violation> violations = List.of(new GetProjectGetLogger(), new GetProject());
```
kelvinou01, line 59: Moved this field to the top of the class

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectGetLoggerTest.java
```diff
@@ -0,0 +1,214 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.guide.errorprone.taskexecution;
+
+import org.junit.jupiter.api.Test;
+
+public class GetProjectGetLoggerTest extends IllegalMethodCalledDuringTaskExecutionTest {
```
kelvinou01, line 21: No changes were done to any tests

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectTest.java
```diff
@@ -0,0 +1,281 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.guide.errorprone.taskexecution;
+
+import org.junit.jupiter.api.Test;
+
+public class GetProjectTest extends IllegalMethodCalledDuringTaskExecutionTest {
```
kelvinou01, line 21: No changes were done to any tests

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -164,61 +155,6 @@ public void fixOrReport(MethodInvocationTree illegalCall, VisitorState state) {
...
-            return false;
-        }
-
-        List<Type> interfaces = ((ClassType) classSym.type).interfaces_field;
-        return interfaces.stream()
-                .anyMatch(ifaceType -> ifaceType.tsym.equals(actionMaybe.get())
-                        && ifaceType.getTypeArguments().size() == 1
-                        && ifaceType.getTypeArguments().get(0).tsym.equals(taskMaybe.get()));
-    }
-
```
kelvinou01, line 221: Moved to `utils.Tasks`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -120,7 +110,7 @@ protected class GetProject implements Violation {
         /** Matches {@code task.getProject()}. */
         @Override
         public boolean matches(MethodInvocationTree illegalCall, VisitorState state) {
-            return TASK_GET_PROJECT.matches(illegalCall, state);
+            return Tasks.TASK_GET_PROJECT.matches(illegalCall, state);
```
felixdesouza: why not make nice util methods like you've done for `isTaskAction` ? I think it reads much better tbh

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/GradleManagedTypeGetPrefix.java
```diff
@@ -57,7 +55,7 @@ public Description matchMethod(MethodTree method, VisitorState state) {
             return Description.NO_MATCH;
         }
 
-        if (!IS_TASK.matches(enclosingClass, state)) {
+        if (!Tasks.TASK.matches(enclosingClass, state)) {
```
felixdesouza: e.g. `Tasks.isTaskClass`

errorprones.md
```diff
@@ -75,7 +75,7 @@ project version as an `@Input public abstract Property<String>`.
 Doing so improves performance in two ways:
 1. It makes your tasks compatible with the configuration cache
 2. It increases task parallelism. When two tasks, such as printProjectName and printProjectVersion, both
-require the same Project object as input, they cannot run in parallel due to prevent concurrent access.
+require the same Project object as input, they cannot run in parallel to prevent concurrent access.
 However, if their inputs are changed to Provider<String> name and Provider<String> version respectively,
```
felixdesouza, line 79:
```suggestion
2. It increases task parallelism. When two tasks, such as `printProjectName` and `printProjectVersion`, both
require the same `Project` object as input, they cannot run in parallel to prevent concurrent access.
However, if their inputs are changed to `Provider<String> name` and `Provider<String> version` respectively,
```

felixdesouza: perhaps in a separate PR or so, but can you go through this page and correctly tag code snippets with ` ? it's actually pretty difficult to read especially if it's read inline.

Note, you should look at the rendered view in your IDE or even in github by clicking the rich diff.

Separately, what's the point of having a `Detailed Link` column, that points to the same thing as the title column?

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectTest.java
felixdesouza, line 21: I'm not a huge fan of abstract test classes tbf, they typically grow unwieldy and hard to change/evolve especially as more abstract descendants are added because it'd then require changing things up and down the hierarchy.

I typically prefer to have smaller scoped helper objects that do the things I want for me. Then you can compose different bits of behaviour by just including the relevant helper objects if need be.

I won't push super hard on it, but largely flagging my distaste for this style of testing

kelvinou01: I think subclassing is the correct model here because there are shared fields + they all fall under 1 bugchecker, but I see where you're coming from. If we want to add additional functionality, let's add it to helper objects rather than bloating `IllegalMethodCalledDuringTaskExecutionTest`.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
kelvinou01: Yep, good idea!

errorprones.md
kelvinou01: Let's FLUP with another PR to fix the error messages. 

Also, I've never noticed that the title column points to the same links as the `Detailed Link` column. I think it should point towards the errorprone implementation instead? I'll FLUP on this as well. Thanks for raising this.

kelvinou01:
FLUPS: 

1. Add `backticks` for code for all BugChecker messages, as errorprones.md is hard to read currently
2. Title column should point towards errorprone implementation
3. Use composition rather than inheritance for testing

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectTest.java
CRogers: I also agree subclassing it not the best here - I would have just made this a helper that you have as a field in each test class and call methods on it. This allows you to compose lots of different helpers - which you can't do with subclassing.

CRogers (approved):

PR #215, opened 2025-09-09
Create a Gradle Errorprone DSL to make adding future task execution violations easy

kelvinou01:
## Before this PR

PRs in order: https://github.com/palantir/gradle-guide/pull/214 ➡️ https://github.com/palantir/gradle-guide/pull/215 (curr) ➡️ https://github.com/palantir/gradle-guide/pull/213 

In the development of errorprones like https://github.com/palantir/gradle-guide/pull/213, quite a lot of logic is duplicated. Let's abstract the general pattern of "match most specific violation --> inject services --> replace illegal method with service method" into a Gradle Errorprone DSL. 

## After this PR

Let's introduce a Gradle Errorprone DSL, to make adding future violations easy. 

==COMMIT_MSG==
Create a Gradle Errorprone DSL to make adding future task execution violations easy
==COMMIT_MSG==


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleService.java
```diff
@@ -0,0 +1,59 @@
...
+package com.palantir.gradle.guide.errorprone.taskexecution;
+
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.suppliers.Supplier;
+import com.sun.tools.javac.code.Type;
+
+/**
+ * Represents a gradle service to be injected as part of an autofix.
+ */
+public class GradleService {
```
CRogers: I'd be tempted to make this an enum as there are only a certain number of `GradleService`s

kelvinou01: Good idea! I've made it an enum.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/TaskExecutionViolation.java
```diff
@@ -0,0 +1,121 @@
...
+import com.sun.source.tree.MethodTree;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
+import java.util.Optional;
+import java.util.Set;
+import one.util.streamex.StreamEx;
+
+/**
+ * Represents strategies to report or auto-fix illegal methods and methods chained to it.
+ */
+public record TaskExecutionViolation(ChainedCall violation, String message, Optional<GradleFix> fix)
```
kelvinou01, line 31: I started writing tests for `TaskExecutionViolation` and `GradleFix` but I don't think testing these in isolation is a good idea. 

We'll have to introduce and maintain a lot of tooling code to wire the tests up (see errorprone's [`SuggestedFixesTest`](https://github.com/google/error-prone/blob/9a2845565ae2820a9245cff26539e51478a3db7f/core/src/test/java/com/google/errorprone/fixes/SuggestedFixesTest.java#L85)). Furthermore, what we care about is the behavior the end-user sees (e.g. `GetProjectTest`, `GetProjectGetProjectDirTest`) and any test cases i'd add to a `TaskExecutionViolationTest` is more succinctly and directly expressed in e.g. `GetProjectGetProjectDirTest` instead. 

All in all, I don't think the marginal  signal we get from testing `TaskExecutionViolation` is worth the maintenance cost. All the cases we care about should be covered in the `<Violation>Test`s anyways.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleService.java
```diff
@@ -0,0 +1,65 @@
...
+package com.palantir.gradle.guide.errorprone.taskexecution;
+
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.suppliers.Supplier;
+import com.sun.tools.javac.code.Type;
+
+/**
+ * Represents a gradle service to be injected as part of an autofix.
+ */
+@SuppressWarnings("ImmutableEnumChecker")
```
kelvinou01, line 26: Because `Supplier` is mutable.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleService.java
```diff
@@ -0,0 +1,65 @@
...
+public enum GradleService {
+    BuildLayout("org.gradle.api.file.BuildLayout", "getBuildLayout"),
+    ProjectLayout("org.gradle.api.file.ProjectLayout", "getProjectLayout"),
+    FileSystemOperations("org.gradle.api.file.FileSystemOperations", "getFileSystemOperations");
```
CRogers: The style in Java is to write enums variants as `SCREAMING_SNAKE_CASE`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/TaskExecutionViolation.java
CRogers: agreed

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/TaskExecutionViolation.java
```diff
@@ -0,0 +1,120 @@
...
+                        .orElse(false);
+                if (canFix) {
+                    descriptionBuilder.addFix(fix.get().render(context, state));
+                }
+            }
+            case TASK -> fix.ifPresent(gradleFix -> descriptionBuilder.addFix(gradleFix.render(context, state)));
+        }
+
+        state.reportMatch(descriptionBuilder.build());
+        return true;
```
CRogers: It feels weird to do this whole part (including getting the method call graph etc) on every single violation. I think I liked it better when there was a `matches` method then you did `fixOrReport` on the first violation that matches. Removes a lot of logic from this class then and I think makes it a bit simpler to work out what's going on.

kelvinou01: yeah this new design does makes it more difficult to reason with performance. Let's separate matching and fixOrReport like before.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleService.java
kelvinou01: Yep. The intention was to make them look similar to the actual service classes to improve readability, but we can use `SCREAMING_SNAKE_CASE` as well!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
```diff
@@ -0,0 +1,111 @@
...
+        boolean isAlreadyAbstract =
+                ASTHelpers.getSymbol(context.enclosingClass.tree()).isAbstract();
+        if (requiresServiceInjection() && !isAlreadyAbstract) {
+            NonAbstractGradleType.maybeTurnClassAbstract(context.enclosingClass.tree(), state)
+                    .ifPresent(fixBuilder::merge);
+        }
+
+        // Inject services
+        for (GradleService service : servicesRequired) {
+            if (!alreadyInjected(service, context.enclosingClass.tree(), state)) {
```
CRogers: I would generally prefer `break`ing earlier here to reduce the nesting. Or using streams:

```java
servicesRequired.stream()
    .filter(service -> !alreadyInjected(...))
    .forEach(
```

CRogers (commented):
LGTM - just one comment to fix then we can merge

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
kelvinou01: Yep!

CRogers (approved):

PR #222, opened 2025-09-15
Add infrastructure to preserve arguments in autofixes

kelvinou01:
## Before this PR

**PRs in sequence**: #222 (curr) -> #223

In the spirit of https://github.com/palantir/gradle-guide/pull/215, let's add the infrastructure necessary to preserve arguments when fixing method calls.

## After this PR

==COMMIT_MSG==
Add infrastructure to preserve arguments in autofixes
==COMMIT_MSG==


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/ChainedCallMatcher.java
```diff
@@ -17,31 +17,32 @@
...
 import java.util.Optional;
 
-public final class ChainedCall {
+public final class ChainedCallMatcher {
     /**
      * To represent {@code foo().bar().baz()}, this list will be [baz, bar, foo].
      * This order facilitates matching in {@code matches}
      */
-    private final List<MethodCall> calls;
+    private final List<Matcher<ExpressionTree>> calls;
```
felixdesouza, line 32: that javadoc is kinda weird, could you not just iterate backwards? not fussed, but mostly curious. Perhaps emphasise that this is in reverse.

nit: `{@code matches}` should be `{@link ChainedCallMatcher#matches}`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/FixTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package com.palantir.gradle.guide.errorprone.taskexecution;
+
+import com.google.common.base.Preconditions;
+
+public final class FixTemplate {
+
+    private String unformatted;
+    private int numFormatArgs;
```
felixdesouza: `final`

felixdesouza: `record` perhaps?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/FixTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private FixTemplate(String unformatted, int numFormatArgs) {
+        Preconditions.checkArgument(
+                0 <= numFormatArgs && numFormatArgs <= 1, "Only 0 or 1 arguments supported at the moment.");
```
felixdesouza: seems overly restrictive, this class seems to work with any number of arguments?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
```diff
@@ -61,17 +63,34 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
 
         SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
 
-        // Replace the illegal call, taking care to preserve any receivers
+        // Preserve arguments (if any)
+        String fixedCallChain;
+        if (fixTemplate.numFormatArgs() == 1) {
+            String args = context.illegalCallToReplace.getArguments().stream()
+                    .map(state::getSourceForNode)
+                    .collect(Collectors.joining(", "));
```
felixdesouza: this seems weird? what are you joining here? should it not be an array?

If so, then it feels like there should be no difference between checking `numFormatArgs == 1` vs not.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/FixTemplate.java
felixdesouza: I imagine to control construction

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
```diff
@@ -61,17 +63,34 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
...
+            fixedCallChain = fixTemplate.render(args);
+        } else {
+            fixedCallChain = fixTemplate.render();
```
felixdesouza: this is also a bit weird, I imagine to deal with the 0 arg case, but that's not what the else statement says here. You're implicitly using the fact that `FixTemplate` doesn't support more than 1 parameter, which is confusing.

It's not clear to me why `FixTemplate` can't support more than 1 parameter. You can still constrain it however you want by having specialised constructors, but have the underlying thing work for all parameters. Sprinkling invariants around feels like an easy to forget thing later.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
felixdesouza: largely commenting without a huge amount of context, but lines 89-93 are the same, but you don't filter out `null` here. Is that something you need to do here?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/FixTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.guide.errorprone.taskexecution;
+
+import com.google.common.base.Preconditions;
+
+public final class FixTemplate {
```
CRogers: I'm a little confused - do we need this approach at all? `Project#copy` has two overrides that take two different args:
1. `Action<? super CopySpec>`
2. `Closure` which takes a `CopySpec`

`FileSystemOperations` has a copy method that takes both of these parameters:
1. Direct overload on `Action<? super CopySpec>`
2. Gradle will automatically create a `Closure` version as it's a managed type.

This means we always we can always go from a `Project#copy` to `FileSystemOperations#copy`.

Do we need any matchers or a fix template type thing? Can't we just always copy the args of the final method call in the chain? For code like `getProject().getProjectDir()` there are no args so it's fine. For this `getProject().copy(...whatever...)` case we just copy the original source code for the args as is - to eg `getFileSystemOperations().copy(...whatever...)`?

CRogers: mmm, or is this because we currently just replace using a string at the moment? I guess some of these cases will  be a bit more difficult, eg there is no exact replacement for `Project#delete` or `Project#mkdir`.

kelvinou01: > Can't we just always copy the args of the final method call in the chain

That would be sufficient for the copy case! But I thought I'd try and implement something which would support mapping an arbitrary number of args to arbitrary locations along the call chain:

`a(argGroupA).b().c(argGroupC).d()` -> `p(argGroupA).q(argGroupC)`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
kelvinou01: Yep, in this block nulls are expected when the illegal chain doesn't have any receivers. e.g getProject() doesn't have a receiver, but task.getProject() does.

```java
        Optional<String> receiverSource =
                Optional.ofNullable(ASTHelpers.getReceiver(innermost)).map(state::getSourceForNode);
        String correctedCallWithReceiver =
                receiverSource.map(receiver -> receiver + ".").orElse("") + fixedCallChain;
        fixBuilder.replace(context.illegalCallToReplace, correctedCallWithReceiver);
```

However, in this block we expect getSourceForNode to never return null. So no null checking is done.

```java
    private String renderFixedCallChain(MethodInvocationTree illegalCall, VisitorState state) {
        if (fixTemplate.numReplacements() == 0) {
            return fixTemplate.render();
        } else if (fixTemplate.numReplacements() == 1) {
            String args = illegalCall.getArguments().stream()
                    .map(state::getSourceForNode)
                    .collect(Collectors.joining(", "));
            return fixTemplate.render(args);
        } else {
            // Additional logic has to be implemented for >1 argument,
            // e.g. From which call in the illegal call chain should the arguments be extracted from?
            // We need something like an ArgumentsMapping — "Arguments for the 2nd call in the illegal chain should
            // go to the 1st call in the fixed chain, while those of the 3rd call should go to the 2nd in the fix."
            throw new IllegalStateException("Logic for multiple replacements not yet supported");
        }
    }
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
kelvinou01: Yep good point, ill make the 0, 1, and unsupported case explicit.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
kelvinou01: From in-person discussion, we join here to create the string to be passed into the FixTemplate

kelvinou01: Also I've updated the variable name of the int field in FixTemplate so hopefully that clears things up

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/FixTemplate.java
kelvinou01: Yep!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
```diff
@@ -117,6 +103,23 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
...
+            String args = illegalCall.getArguments().stream()
+                    .map(state::getSourceForNode)
+                    .collect(Collectors.joining(", "));
+            return fixTemplate.render(args);
+        } else {
+            // Additional logic has to be implemented for >1 argument,
+            // e.g. From which call in the illegal call chain should the arguments be extracted from?
+            // We need something like an ArgumentsMapping — "Arguments for the 2nd call in the illegal chain should
+            // go to the 1st call in the fixed chain, while those of the 3rd call should go to the 2nd in the fix."
+            throw new IllegalStateException("Logic for multiple replacements not yet supported");
```
felixdesouza: I thought we weren't doing any argument mapping anymore? 

plus I think your docs are still muddled, are you referring to arguments or replacements?

is the idea that if you had a `FixTemplate` with more than one replacement, you'd have multiple `MethodInvocationTree illegalCall`s ?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/TaskExecutionViolation.java
```diff
@@ -28,7 +28,7 @@
 /**
  * Represents strategies to report or auto-fix illegal methods and methods chained to it.
  */
-public record TaskExecutionViolation(ChainedCallMatcher chainedCallMatcher, String message, Optional<GradleFix> fix)
+public record TaskExecutionViolation(ChainedCallMatcher violationMatcher, String message, Optional<GradleFix> fix)
```
felixdesouza, line 31: did you mean to keep the class name the same?

kelvinou01: yep.

kelvinou01: ChainedCallMatcher is a general matching mechanism (like an extension of Matcher<ExpressionTree>), while violationMatcher is the specific ChainedCallMatcher used to match violations.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
kelvinou01: Let's ditch the argument mapping notion entirely from the code. I've made `FixTemplate` a sealed interface. Let me know what you think.

kelvinou01: > plus I think your docs are still muddled, are you referring to arguments or replacements?

yep i've unified the terminology, i meant replacements, thanks for raising this

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
```diff
@@ -98,6 +103,19 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
...
+        if (fixTemplate instanceof FixTemplate.Nullary nullary) {
+            return nullary.fix();
+        } else if (fixTemplate instanceof FixTemplate.Unary unary) {
+            String args = illegalCall.getArguments().stream()
+                    .map(state::getSourceForNode)
+                    .collect(Collectors.joining(", "));
+            return unary.render(args);
+        } else {
+            throw new IllegalStateException("Unknown fix template type: " + fixTemplate);
+        }
```
kelvinou01: Sadly no switching on sealed interfaces in java 17 :(

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/Replacement.java
```diff
@@ -0,0 +1,63 @@
...
+
+    record TemplatedReplacement(String unformatted) implements Replacement {
+        @Override
+        public String render(MethodInvocationTree tree, VisitorState state) {
+            Preconditions.checkArgument(!tree.getArguments().isEmpty());
+            String args =
+                    tree.getArguments().stream().map(state::getSourceForNode).collect(Collectors.joining(", "));
+            return String.format(unformatted, args);
+        }
+    }
```
kelvinou01: Example: https://github.com/palantir/gradle-guide/pull/223/files#diff-8afdfa5877d127578d24f129533fe7e3f803bdaaf003d4dc8d29375600e6e433

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
```diff
@@ -36,15 +36,15 @@
  * An autofix for a Gradle task. This usually involves injecting a Gradle service, and replacing violating methods
  * with methods from the service.
  * @param servicesRequired The Gradle services required to make this fix
- * @param correctedCall Replaces the violating chained call
+ * @param fixTemplate Replaces the violating chained call
  */
-public record GradleFix(List<GradleService> servicesRequired, String correctedCall) {
-    public static GradleFix of(GradleService service, String correctCall) {
-        return new GradleFix(List.of(service), correctCall);
+public record GradleFix(List<GradleService> servicesRequired, Replacement fixTemplate) {
```
felixdesouza: name change from `fixTemplate` -> `replacement` ?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/Replacement.java
```diff
@@ -0,0 +1,63 @@
...
+
+    record LiteralReplacement(String fix) implements Replacement {
+        @Override
+        public String render(MethodInvocationTree tree, VisitorState state) {
+            Preconditions.checkArgument(tree.getArguments().isEmpty());
+            return fix;
+        }
+    }
+
+    record TemplatedReplacement(String unformatted) implements Replacement {
```
felixdesouza: nit: `unformatted` -> `template`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/TaskExecutionViolation.java
```diff
@@ -28,25 +28,25 @@
 /**
  * Represents strategies to report or auto-fix illegal methods and methods chained to it.
  */
-public record TaskExecutionViolation(ChainedCall violation, String message, Optional<GradleFix> fix)
+public record TaskExecutionViolation(ChainedCallMatcher violationMatcher, String message, Optional<GradleFix> fix)
         implements Comparable<TaskExecutionViolation> {
 
-    public static TaskExecutionViolation fix(ChainedCall violation, String message, GradleFix fix) {
+    public static TaskExecutionViolation fix(ChainedCallMatcher violation, String message, GradleFix fix) {
```
felixdesouza: nit, consistent name of `violation` or `violationMatcher` across this file. Not fussed which one you pick

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/GradleFix.java
```diff
@@ -36,15 +36,15 @@
...
+    public static GradleFix of(GradleService service, Replacement fixedCall) {
+        return new GradleFix(List.of(service), fixedCall);
     }
 
-    public static GradleFix of(String correctCall) {
-        return new GradleFix(List.of(), correctCall);
+    public static GradleFix of(Replacement fixedCall) {
+        return new GradleFix(List.of(), fixedCall);
```
felixdesouza: `fixedCall` -> `replacement`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/ChainedCallMatcher.java
```diff
@@ -17,31 +17,32 @@
...
 import java.util.List;
 import java.util.Optional;
 
-public final class ChainedCall {
+public final class ChainedCallMatcher {
     /**
-     * To represent {@code foo().bar().baz()}, this list will be [baz, bar, foo].
-     * This order facilitates matching in {@code matches}
+     * To represent {@code foo().bar().baz()}, this list will be in reverse — [baz, bar, foo].
+     * This order facilitates matching in {@code ChainedCallMatcher#matches}
```
felixdesouza:
```suggestion
     * This order facilitates matching in {@link ChainedCallMatcher#matches}
```

`@code` is for code blocks, but does not enable linking i.e. good for examples
`@link` is an actual link

kelvinou01: gotcha, thanks for the tip!

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/TaskExecutionViolation.java
kelvinou01: yep!

felixdesouza (approved):

PR #223, opened 2025-09-15
Autofix `getProject().copy()` to `getFileSystemOperations().copy()`

kelvinou01:
## Before this PR

**PRs in sequence**: #222 -> #223 (curr)

We've added an [errorprone](https://github.com/palantir/gradle-guide/pull/142) to prevent people from using `getProject` during the execution time. In most cases, when people do `getProject().X()` during execution time,  we can substitute it for another method which is configuration cache friendly — see these [metrics](https://pl.ntr/2vj). `getProject().copy()` is one such case, let's autofix these!

## After this PR

==COMMIT_MSG==
Autofix `getProject().copy()` to `getFileSystemOperations().copy()`
==COMMIT_MSG==

### DIFF
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
@@ -69,6 +69,10 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
             .onDescendantOf("org.gradle.api.Project")
             .named("getProjectDir")
             .withNoParameters();
+    private static final Matcher<ExpressionTree> copy = Matchers.instanceMethod()
+            .onDescendantOf("org.gradle.api.Project")
+            .named("copy")
+            .withParameters("org.gradle.api.Action");
 
     private static final TaskExecutionViolation REPORT_GET_PROJECT = TaskExecutionViolation.report(
             ChainedCallMatcher.of(getProject), "Don't call `getProject()` in task actions");
@@ -83,9 +87,17 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
             GradleFix.of(
                     GradleService.PROJECT_LAYOUT,
                     Replacement.literal("getProjectLayout().getProjectDirectory().getAsFile()")));
+    private static final TaskExecutionViolation FIX_GET_PROJECT_COPY = TaskExecutionViolation.fix(
+            ChainedCallMatcher.of(getProject, copy),
+            "Instead of `getProject().copy(...)`, do `getFileSystemOperations().copy(...)`",
+            GradleFix.of(
+                    GradleService.FILE_SYSTEMS_OPERATIONS, Replacement.template("getFileSystemOperations().copy(%s)")));
 
     private static final List<TaskExecutionViolation> violationsInOrderOfSpecificity = Stream.of(
-                    FIX_GET_PROJECT_GET_PROJECT_DIR, FIX_GET_PROJECT_GET_LOGGER, REPORT_GET_PROJECT)
+                    FIX_GET_PROJECT_COPY,
+                    FIX_GET_PROJECT_GET_PROJECT_DIR,
+                    FIX_GET_PROJECT_GET_LOGGER,
+                    REPORT_GET_PROJECT)
             .sorted()
             .toList();
 
--- a/gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectCopyTest.java
+++ b/gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectCopyTest.java
@@ -1,0 +1,176 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -83,9 +87,17 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
             GradleFix.of(
                     GradleService.PROJECT_LAYOUT,
                     Replacement.literal("getProjectLayout().getProjectDirectory().getAsFile()")));
+    private static final TaskExecutionViolation FIX_GET_PROJECT_COPY = TaskExecutionViolation.fix(
+            ChainedCallMatcher.of(getProject, copy),
+            "Instead of `getProject().copy(...)`, do `getFileSystemOperations().copy(...)`",
+            GradleFix.of(
+                    GradleService.FILE_SYSTEMS_OPERATIONS, Replacement.template("getFileSystemOperations().copy(%s)")));
```
CRogers, line 94: This is nice!

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectCopyTest.java
```diff
@@ -0,0 +1,176 @@
...
+                import javax.inject.Inject;
+                import org.gradle.api.DefaultTask;
+                import org.gradle.api.Plugin;
+                import org.gradle.api.Project;
+                import org.gradle.api.file.FileSystemOperations;
+                import org.gradle.api.tasks.TaskAction;
+                abstract class AlreadyInjectedTask extends DefaultTask {
+                    private static final String IM_AT_THE_TOP_OF_THE_CLASS = "happy_squirrel.txt";
+                    @Inject
+                    protected abstract FileSystemOperations getFileSystemOperations();
```
felixdesouza, line 80: I'd be curious to see if you had a test where you already had a `FileSystemOperations` but named something else.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
```diff
@@ -83,9 +87,17 @@ public final class IllegalMethodCalledDuringTaskExecution extends GradleGuideBug
             GradleFix.of(
                     GradleService.PROJECT_LAYOUT,
                     Replacement.literal("getProjectLayout().getProjectDirectory().getAsFile()")));
+    private static final TaskExecutionViolation FIX_GET_PROJECT_COPY = TaskExecutionViolation.fix(
+            ChainedCallMatcher.of(getProject, copy),
+            "Instead of `getProject().copy(...)`, do `getFileSystemOperations().copy(...)`",
```
felixdesouza, line 92: I think I'm leaning towards the javadoc identifier mechanism so it's not tied to whatever you've chosen to name the method i.e. `FileSystemOperations#copy` and you know what method on what type.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/taskexecution/IllegalMethodCalledDuringTaskExecution.java
felixdesouza: yes!

felixdesouza, line 94: okay, thinking about this a little bit more, there's nothing that binds `GradleService.FILE_SYSTEM_OPERATIONS` to the constant lurking inside your template of `getFileSystemOperations().copy(%s)`. 

This is separate to this PR and probably a worthwhile FLUP.

I think the relationship between `GradleService`, `GradleFix` and `Replacement` should be inverted:

* The replacement is effectively what defines what services need to be injected
* You also bind the getter that gets rendered for the service injection to whatever is in the template
* I imagine most replacements only require a single service? 

What if you could define the intent of what you want to do from the service where you want to make it?

i.e. something like:

```java
private static final TaskExecutionViolation FIX_GET_PROJECT_COPY = TaskExecutionViolation.fix(
            ChainedCallMatcher.of(getProject, copy),
            "Instead of `getProject().copy(...)`, do `getFileSystemOperations().copy(...)`", 1Code has comments. Press enter to view.
            GradleFix.of(
                    GradleService.FILE_SYSTEMS_OPERATIONS.copy()));
```

you can think of a `Replacement` as potentially having some sort of service dependency. and `GradleService.FILE_SYSTEMS_OPERATIONS.copy()` returning a `Replacement` with the right bindings.

Also, I think it will fix the issue I stated in a comment in your test about a pre-existing `FileSystemOperations` that's named slightly differently. You can then bind to that name instead and fallback to the default name if it's not been injected already.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectCopyTest.java
CRogers: It's not great but I think you can just inject the same service multiple times

felixdesouza (approved):
Looks good!

Let's FLUP on the binding piece, so we don't introduce multiple instances of the same thing.

PR #235, opened 2025-10-02
Refactor `GradleFix`, `GradleService`, and `Replacement` to bind each fix type to a gradle service

kelvinou01:
## Before this PR

Although it is implicit that a fix to task execution violation is always bound to zero or one gradle services (and involves injecting said service), our types didn't reflect this.

## After this PR

Refactor `GradleFix`, `GradleService`, and `Replacement` to bind each fix type to a gradle service. This deduplicates the code to instantiate services. 

==COMMIT_MSG==
Refactor `GradleFix`, `GradleService`, and `Replacement` to bind each fix type to a gradle service
==COMMIT_MSG==


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/GradleFix.java
```diff
@@ -29,26 +31,29 @@
...
 
 /**
  * An autofix for a Gradle task. This usually involves injecting a Gradle service, and replacing violating methods
  * with methods from the service.
- * @param servicesRequired The Gradle services required to make this fix
+ * @param serviceMaybe The Gradle services required to make this fix
  * @param replacement Replaces the violating chained call
  */
-public record GradleFix(List<GradleService> servicesRequired, Replacement replacement) {
+public record GradleFix(Optional<GradleService> serviceMaybe, Replacement replacement) {
```
kelvinou01: Moved these types to `types` to clarify that these types shouldn't be coupled with task execution. 

We want to reuse these types to fix things at e.g. `Plugin#apply` time as well.

gradle-guide-error-prone/src/test/java/com/palantir/gradle/guide/errorprone/taskexecution/GetProjectCopyTest.java
```diff
@@ -108,6 +108,104 @@ final void action() {
                     }
                 }
                 """);
+
+        testFix(
```
kelvinou01, line 112: Thought of creating new test methods for these but these actually test variations of the same specified behavior. 

Also as @felixdesouza pointed out, the error messages outputted by error-prone tests is not ideal. Ideally the computer diffs the thing for us.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/GradleFix.java
```diff
@@ -61,8 +66,11 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
 
         SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
 
+        // Either getService() method or service field
```
felixdesouza: nit:
```suggestion
        // Either service method getter or field
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/GradleFix.java
```diff
@@ -61,8 +66,11 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
...
+        String receiver = getServiceReceiver(context, state, fixBuilder);
```
felixdesouza: why not name this `serviceReceiver` so you're not having to do the weird `recv` further down

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/GradleFix.java
```diff
@@ -83,32 +91,47 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
...
     }
 
     @SuppressWarnings("ASTHelpersSuggestions")
-    private boolean alreadyInjected(GradleService service, ClassTree classTree, VisitorState state) {
-        return ASTHelpers.getSymbol(classTree).getEnclosedElements().stream()
+    private Optional<String> findServiceReceiver(GradleService service, ClassTree classTree, VisitorState state) {
+        List<Symbol> enclosedElements = ASTHelpers.getSymbol(classTree).getEnclosedElements();
+        Stream<String> fields = enclosedElements.stream()
+                .filter(symbol -> symbol.getKind().equals(ElementKind.FIELD))
+                .filter(symbol -> ASTHelpers.isSameType(symbol.type, service.getType(state), state))
```
felixdesouza: let's use `asType()`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/Replacement.java
```diff
@@ -14,48 +14,67 @@
...
             String args =
                     tree.getArguments().stream().map(state::getSourceForNode).collect(Collectors.joining(", "));
             return String.format(template, args);
         }
     }
 
-    String render(MethodInvocationTree tree, VisitorState state);
+    protected abstract String renderCall(MethodInvocationTree tree, VisitorState state);
+
+    final String render(String receiver, MethodInvocationTree tree, VisitorState state) {
```
felixdesouza: yeah this whole bit is weird, it doesn't do anything with the receiver except render. The underlying impls don't know about it. Constructors don't know about it. We should just move the receiver concat code into `GradleFix`, it just looks like it doesn't belong in here.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/GradleFix.java
```diff
@@ -83,32 +91,47 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
...
+        if (serviceMaybe.isEmpty()) {
+            // This means the method is on DefaultTask (i.e. `this`). No receiver needed.
+            return "";
+        }
+        GradleService service = serviceMaybe.get();
+        return findServiceReceiver(service, context.enclosingClass.tree(), state)
+                .orElseGet(() -> {
+                    injectService(context, state, service, fixBuilder);
+                    return service.getterName() + "()";
                 });
```
felixdesouza: I kinda hate this chaining, especially considering there's a side-effect in here. Can we try:

```java
Optional<String> serviceReceiver = findServiceReceiver(service, context.enclosingClass.tree(), state)
if (serviceReceiver.isPresent()) {
  return serviceReceiver.get();
}

injectService(context, state, service, fixBuilder);
return service.getterName() + "()"
```

felixdesouza: `getterName` - >`defaultGetterName` or `newGetterName` something along those lines

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/GradleFix.java
```diff
@@ -83,32 +91,47 @@ public SuggestedFix render(GradleFixContext context, VisitorState state) {
...
-                            String.format(
-                                    "@Inject\nprotected abstract %s %s();", service.className(), service.getterName()));
-                    fixBuilder.merge(serviceGetter);
+    private String getServiceReceiver(GradleFixContext context, VisitorState state, Builder fixBuilder) {
+        if (serviceMaybe.isEmpty()) {
+            // This means the method is on DefaultTask (i.e. `this`). No receiver needed.
+            return "";
+        }
+        GradleService service = serviceMaybe.get();
+        return findServiceReceiver(service, context.enclosingClass.tree(), state)
```
felixdesouza: Can we reference the method instead of the field?

```suggestion
        return findServiceReceiver(service, context.enclosingClass().tree(), state)
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/types/GradleFix.java
```diff
@@ -29,26 +31,29 @@
...
 import java.util.Optional;
+import java.util.stream.Stream;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.Name;
 
 /**
  * An autofix for a Gradle task. This usually involves injecting a Gradle service, and replacing violating methods
  * with methods from the service.
- * @param servicesRequired The Gradle services required to make this fix
+ * @param serviceMaybe The Gradle services required to make this fix
```
felixdesouza: let's flip this `maybeService`

or wild idea, just have it be `service`. I think it reads better imo:

```java
if (service.isPresent()) {
  // do bla
}
```

felixdesouza (dismissed):
let's get this in! I think there shouldn't be too many changes to the infra now 🙏  for the rest of the changes you gotta get in

esword:
👍

PR #236, opened 2025-10-03
Ban `task1.dependsOn(task2)`, unless if we think `task1` is a lifecycle task

kelvinou01:
## Before this PR

`task1.dependsOn(task2)` is a sign of bad task design, and can cause tasks to be run when unnecessary. Lets ban it.

## After this PR

==COMMIT_MSG==
Ban `task1.dependsOn(task2)`, unless if we think task1 is a lifecycle task
==COMMIT_MSG==

### DIFF
--- a/errorprones.md
+++ b/errorprones.md
@@ -110,7 +110,7 @@ When defining a custom Task or Extension, you should make it an abstract class w
 </td>
 <td>
 
-Instead of `task1.dependsOn(task2)`, wire up the outputs of task2 to the inputs of task1 using providers.
+Instead of `task1.dependsOn(task2)`, wire up the outputs of `task2` to the inputs of `task1` using providers.
 Using `dependsOn` makes it easy to add unnecessary task dependencies — e.g. depending on `jar` when you just
 need `classes`. It is also a sign of bad task design. You should write tasks with explicit, fine-grained inputs
 and outputs. Having unnecessary dependencies adds unnecessary work to a build and hurts task parallelism.
--- a/errorprones.md
+++ b/errorprones.md
@@ -98,5 +98,22 @@ When defining a custom Task or Extension, you should make it an abstract class w
 
 </td>
 </tr>
+
+<tr>
+<td>
+
+<a id="TaskDependsOn" href="guide/diagnosing-build-performance.md#needlessly-included-tasks">`TaskDependsOn`</a>
+
+</td>
+<td>
+<a href="guide/diagnosing-build-performance.md#needlessly-included-tasks">Please read</a>
+</td>
+<td>
+
+Instead of `task1.dependsOn(task2)`, wire up the outputs of task2 to the inputs of task1 using providers.
+
+
+</td>
+</tr>
 </tbody>
 </table>
--- a/errorprones.md
+++ b/errorprones.md
@@ -112,8 +112,8 @@ When defining a custom Task or Extension, you should make it an abstract class w
 
 Instead of `task1.dependsOn(task2)`, wire up the outputs of task2 to the inputs of task1 using providers.
 Using `dependsOn` makes it easy to add unnecessary task dependencies — e.g. depending on `jar` when you just
-need `classes`. It also doesn't encourage you to write tasks with explicit, fine-grained inputs and outputs.
-Having unnecessary dependencies adds unnecessary work to a build and hurts task parallelism.
+need `classes`. It is also a sign of bad task design. You should write tasks with explicit, fine-grained inputs
+and outputs. Having unnecessary dependencies adds unnecessary work to a build and hurts task parallelism.
 
 However, there are cases where `dependsOn` is unavoidable:
 1. Wiring up tasks to lifecycle tasks e.g. `build.dependsOn(myTask)`
--- a/errorprones.md
+++ b/errorprones.md
@@ -111,6 +111,17 @@ When defining a custom Task or Extension, you should make it an abstract class w
 <td>
 
 Instead of `task1.dependsOn(task2)`, wire up the outputs of task2 to the inputs of task1 using providers.
+Using `dependsOn` makes it easy to add unnecessary task dependencies — e.g. depending on `jar` when you just
+need `classes`. It also doesn't encourage you to write tasks with explicit, fine-grained inputs and outputs.
+Having unnecessary dependencies adds unnecessary work to a build and hurts task parallelism.
+
+However, there are cases where `dependsOn` is unavoidable:
+1. Wiring up tasks to lifecycle tasks e.g. `build.dependsOn(myTask)`
+2. Tasks that do system-wide setup like installing npm. In these cases, we recommended that you specify the
+    installation directory as an `@OutputDirectory` of the task anyways.
+
+This errorprone tries to detect legitimate uses of `dependsOn`. You can suppress false positives, but we
+encourage you to do so sparingly, and think about whether something can be improved about your task design.
 
 
 </td>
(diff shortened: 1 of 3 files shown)


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
```diff
@@ -0,0 +1,135 @@
...
+        1. Wiring up tasks to lifecycle tasks e.g. `build.dependsOn(myTask)`
+        2. Tasks that do system-wide setup like installing npm. In these cases, we recommended that you specify the
+            installation directory as an `@OutputDirectory` of the task anyways.
+
+        This errorprone tries to detect legitimate uses of `dependsOn`. You can suppress false positives, but we
+        encourage you to do so sparingly, and think about whether something can be improved about your task design.
+        """)
+public final class TaskDependsOn extends GradleGuideBugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final Set<String> KNOWN_LIFECYCLE_TASKS = Set.of(
```
felixdesouza: lifecycle tasks typically do not have any implementation i.e. there are just arbitrary tasks with no task actions, but real tasks attach to them i.e. `check`, `build` etc.

felixdesouza: still reading through the PR, perhaps it's a matter of nomenclature, but will see shortly

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (!TASK_DEPENDS_ON.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        Optional<ExpressionTree> receiverMaybe = Optional.ofNullable(ASTHelpers.getReceiver(tree));
+        if (receiverMaybe.isEmpty()) {
+            return Description.NO_MATCH;
+        }
```
felixdesouza, line 66: this will ignore anything that's written in a task constructor, but not sure how many of those things exist

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
```diff
@@ -0,0 +1,135 @@
...
+        ExpressionTree receiver = receiverMaybe.get();
+        Type taskType = ASTHelpers.getType(receiver);
+
+        // It's alright to wire up lifecycle tasks with dependsOn
+        if (isKnownLifecycleTask(taskType, state)) {
+            return Description.NO_MATCH;
+        }
```
felixdesouza: yeah, given the impl, this is just not true

e.g. for a lot of cases, it's weird that `jar` has a depends on, it means that something has gone wrong else where.

It kinda sounds like you're conflating tasks for implementations you don't control, and the actual meaning of lifecycle tasks.

From [Organizing tasks](https://docs.gradle.org/current/userguide/organizing_tasks.html):

>Actionable tasks in Gradle are tasks that perform actual work, such as compiling code. Lifecycle tasks are tasks that do not do work themselves. These tasks have no actions, instead, they bundle actionable tasks and serve as targets for the build.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
```diff
@@ -0,0 +1,135 @@
...
+        // We assume people don't write custom types for lifecycle tasks
+        if (!isCustomTask(taskType, state)) {
+            return Description.NO_MATCH;
+        }
```
felixdesouza: What example does this cover?

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
kelvinou01: yep you're right, I was conflating the two. lets remove everythng that has to do with `KNOWN_LIFECYCLE_TASK` here entirely.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
kelvinou01: e.g. `check` is a `DefaultTask`

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
```diff
@@ -0,0 +1,90 @@
...
+import com.sun.tools.javac.code.Type;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = SeverityLevel.ERROR,
+        summary =
+                """
+        Instead of `task1.dependsOn(task2)`, wire up the outputs of task2 to the inputs of task1 using providers.
+        Using `dependsOn` makes it easy to add unnecessary task dependencies — e.g. depending on `jar` when you just
```
felixdesouza:
```suggestion
        Instead of `task1.dependsOn(task2)`, wire up the outputs of `task2` to the inputs of `task1` using providers.
        Using `dependsOn` makes it easy to add unnecessary task dependencies — e.g. depending on `jar` when you just
```

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/TaskDependsOn.java
```diff
@@ -0,0 +1,90 @@
...
+        ExpressionTree receiver = receiverMaybe.get();
+        Type taskType = ASTHelpers.getType(receiver);
+
+        if (!isCustomTask(taskType, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree)
+                .setMessage("Instead of `task1.dependsOn(task2)`, "
+                        + "wire up the outputs of task2 to the inputs of task1 using providers")
```
felixdesouza:
```suggestion
                        + "wire up the outputs of `task2` to the inputs of `task1` using providers")
```

errorprones.md
```diff
@@ -96,6 +96,34 @@ the tasks become independent and can execute in parallel.
...
+
+<a id="TaskDependsOn" href="guide/diagnosing-build-performance.md#needlessly-included-tasks">`TaskDependsOn`</a>
+
+</td>
+<td>
+<a href="guide/diagnosing-build-performance.md#needlessly-included-tasks">Please read</a>
+</td>
+<td>
+
+Instead of `task1.dependsOn(task2)`, wire up the outputs of task2 to the inputs of task1 using providers.
```
felixdesouza:
```suggestion
Instead of `task1.dependsOn(task2)`, wire up the outputs of `task2` to the inputs of `task1` using providers.
```

kelvinou01:
Not blocking, and unsure about all this:

While errorprones like this one provide ratchet to discourage new usages of dependsOn, it feels like most devs don't know what they mean, and that it's not easy to find out what they mean. 

Maybe we can improve on how suppressions signal the intent of the bugchecker to devs. Reflecting about my own workflow: When I encounter @SuppressWarnings("UnusedVariable"), I can do an exact text search and get to errorprone.info/bugpattern/UnusedVariable. For @SuppressWarnings("LineLength"), I can find  checkstyle.sourceforge.io/checks/sizes/linelength.html. But there's no easily google-able thing for gradle-guide. When I can't find anything on google, I probably just give up and move on with my day.

Would `@BadGradle("TaskDependsOn")` be higher signal? It encapsulates the intent behind the suppression (bad gradle), and stands out from the crowd of other suppressions which could spark curiosity to find out more.  Maybe it comes across as shaming and unpleasant.

FinlayRJW:
👍

PR #256, opened 2025-10-20
Add `UseGradleLoggerInGradlePlugins` to stop people from using `SafeLogger` or generic slf4j `Logger`s within Gradle plugins

kelvinou01:
## Before this PR

We should use Gradle's logger instead of SafeLogger in gradle 

- Gradle’s type exposes the lifecycle log level which will always print the log but does not colour it as a warn or error. Sometimes useful.
- [SafeLogger infra](https://github.com/palantir/safe-logging) outputs logs as json. This is very bad for Gradle as we have people reading the logs, not machines
- There is no need to have the overhead of safe vs unsafe args for our internal developments. Everything should be safe

## After this PR

Let's
- Autofix generic slf4j loggers to gradle loggers
- Flag SafeLogger usage. We don't autofix because the cost of autofixing somewhere we shouldn't is very large — logging sensitive information in an unsafe log.

==COMMIT_MSG==
Add `UseGradleLoggerInGradlePlugins` to stop people from using `SafeLogger` or generic slf4j `Logger`s within Gradle plugins
==COMMIT_MSG==

### DIFF
--- a/gradle-guide-error-prone/build.gradle
+++ b/gradle-guide-error-prone/build.gradle
@@ -14,6 +14,7 @@ dependencies {
     testImplementation 'org.assertj:assertj-core'
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'com.google.errorprone:error_prone_test_helpers'
+    testImplementation 'com.palantir.safe-logging:logger'
     testImplementation gradleApi()
 }
 
@@ -36,4 +37,8 @@ moduleJvmArgs {
         'jdk.compiler/com.sun.tools.javac.util'
     ]
     opens = ['jdk.compiler/com.sun.tools.javac.comp']
+}
+
+checkUnusedDependencies {
+    ignore('com.palantir.safe-logging', 'logger')  // Used within a test fixture
 }
--- a/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/UseGradleLoggerInGradlePlugins.java
+++ b/gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/UseGradleLoggerInGradlePlugins.java
@@ -1,0 +1,99 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.guide.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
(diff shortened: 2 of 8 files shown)


gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/UseGradleLoggerInGradlePlugins.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.errorprone.matchers.Matchers;
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.VariableTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = SeverityLevel.ERROR,
+        summary = "Use Gradle's `Logger` instead of `SafeLogger` within Gradle plugins. "
+                + "Safe logging isn't necessary for Gradle plugins which don't process sensitive data.")
+public final class UseGradleLoggerInGradlePlugins extends GradleGuideBugChecker
```
CRogers, line 39: We should also force direct slf4j loggers (ie `private static final org.slf4j.logger.Logger log`) to be the Gradle logging type too. This should be safe to autofix. There are far, far more instances of this than the SafeLogger one, but the safe logger one does cause more issues.

CRogers: Can be done in a follow up PR.

gradle-guide/src/main/java/com/palantir/gradle/guide/GradleGuidePlugin.java
```diff
@@ -28,7 +28,8 @@ public class GradleGuidePlugin implements Plugin<Project> {
             "ConfigurationAvoidanceRegistration",
             "ProviderGet",
             "NonAbstractGradleType",
-            "IllegalMethodCalledDuringTaskExecution");
+            "IllegalMethodCalledDuringTaskExecution",
+            "UseGradleLoggerInGradlePlugins");
```
CRogers, line 32: This doesn't actually have any fixes, but I guess no harm in putting here for now. The ideal would be a "include everything in package" option in the suppressible-error-prone extension.

gradle-guide-error-prone/build.gradle
```diff
@@ -37,3 +38,7 @@ moduleJvmArgs {
     ]
     opens = ['jdk.compiler/com.sun.tools.javac.comp']
 }
+
+checkUnusedDependencies {
+    ignore('com.palantir.safe-logging', 'logger')  // Used within a test fixture
```
CRogers, line 43: You _might_ be able to avoid this by setting safe-logging to be `testCompileOnly` instead of `testImplementation`.

gradle-guide-error-prone/src/main/java/com/palantir/gradle/guide/errorprone/UseGradleLoggerInGradlePlugins.java
kelvinou01: I see! I didn't know that `org.slf4j.logger.Logger` instances was an issue as well. Let's autofix in this PR

gradle-guide-error-prone/build.gradle
kelvinou01: I tried that but `flag_safe_logger_instances` fails with:

```
java.lang.AssertionError: Saw unexpected error on line 3. All errors:
/Test.java:3: error: package com.palantir.logsafe.logger does not exist
import com.palantir.logsafe.logger.SafeLogger;
```

Would be nice if CheckUnusedDependenciesTask had a way of checking if something was used in a test fixture. I don't think it's possible without some hackery?

FinlayRJW:
👍

FinlayRJW:
👍

FinlayRJW:
👍

PR #423, opened 2026-05-11
Switch from deprecated `AST_MATCH` to `TEXT_MATCH` mode

kelvinou01:
## Before this PR

error-prone 2.49.0 switches `BugCheckerRefactoringTestHelper` to [TEXT_MATCH mode as default](https://github.com/google/error-prone/commit/ccfc176eb7#diff-d1a322c5302810becfb0cbc80fb0d611800a40b3682ad149b8bb2d08623d0c33R288) and [deprecates AST_MATCH mode](https://github.com/google/error-prone/commit/190e452caa94e7a351e4e6f585fc1d7a640c6f1a). 

This tightens the criterion for an (actual, expected) pair to be matched. Whitespaces and comments are taken into consideration now.

This is a purely test behavior change, no consumers should be expected.

## After this PR

Switch to TEXT_MATCH mode in preparation for error-prone 2.49.0


FinlayRJW:
👍

